PR #38, opened 2024-08-06
Fix assorted spelling errors spotted with codespell

peterjc:
See https://github.com/codespell-project/codespell

Also fixes a couple of docstring style issues, D205 and D401, but does not resolve the outstanding rust lint errors (mainly S603 and S607).


peterjc:
I would propose to add codespell to the pre-commit configuration next, see e.g. these setups which include a few exclusions (typically abbreviations or biological words which are similar to an English word):

https://github.com/peterjc/thapbi-pict/blob/v1.0.13/.pre-commit-config.yaml#L94
https://github.com/biopython/biopython/blob/biopython-184/.pre-commit-config.yaml#L76

widdowquinn:
> I would propose to add codespell to the pre-commit configuration next

Seems reasonable to me - go ahead.

widdowquinn (approved):

Issue #32, opened 2024-07-31
Add module that checks for complete paths to executables


widdowquinn:
## Summary:

For a range of reasons, including security and good practice/logging, it is desirable to have a module that, for each third-party application which will be called from the command-line, checks for a viable `$PATH` to that application and makes it available within pyani-plus.

The internal variables providing these paths can then be used in, for example, `subprocess.run()` commands in target fixture-generating scripts. If no `$PATH` is found we can prompt the user to install a missing tool.

## Description:

See [`dependencies.py`](https://github.com/widdowquinn/pyani/blob/master/pyani/dependencies.py) and related code in the original `pyani` repo for examples of how this can be done.

What I'm envisioning ATM is a module, e.g. `pyani-plus.dependencies` where we can get the path to - say - `nucmer` with something like

```python
from pyani-plus import dependencies

print(dependencies.nucmer_path)
print(dependencies.nucmer_version)
```

and have the underlying code return the appropriate `Path` or `str` object. Then, when running `nucmer` we'd replace the string that _assumes_ nucmer is in the `$PATH` with, say the `nucmer_path` variable. I think this should also apply within the snakemake workflows, but I'm not 100% sure how that looks, right now.

peterjc:
See also https://github.com/peterjc/thapbi-pict/blob/master/thapbi_pict/versions.py where I have snippets of code to get the version of command line tools (e.g. https://github.com/peterjc/thapbi-pict/blob/v1.0.13/thapbi_pict/versions.py handles NCBI BLAST+ binaries).

PR #39, opened 2024-08-06
Fix or ignore remaining ruff checks

peterjc:
This is some minor housekeeping to resolve the currently failing pre-commit checks.

Adds local ignores for S607 pending a proper fix.

Closes #36, cherry-picking the pre-commit tool updates.

Given the existence of conflicting and deprecated rules in ruff check, there ought to be a better approach to using ruff with all rules...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.4.1
+  rev: v0.5.6
   hooks:
     # Run the linter.
     - id: ruff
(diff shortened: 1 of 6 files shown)


widdowquinn:
> Adds local ignores for S607 pending a proper fix.

Linking #32 as this is the suggested long-term fix and should allow us to remove any local instructions to ignore the rule.

widdowquinn (commented):
I quite like COM812 but not enough to object.

widdowquinn (commented):
Should `indir` be a `Path` in b4d56485 l.46?

peterjc:
Given the code used ``Path(indir)`` my inference was ``indir`` was assumed to be a string. Perhaps it can be either?

(In fact, I'm not sure the function is currently used at all)

peterjc:
I don't immediately see why COM812 for trailing commas should clash with ISC001 (implicit string concatenation), but personally I'd rather have the latter checked. I think the ruff/black formater itself does a sensible job with trailing commas anyway.

peterjc:
Other the codecov, the tests are all green - OK to merge this now?

widdowquinn:
Brilliant - thanks @peterjc !

PR #43, opened 2024-08-08
BioConda mummer package fixed

peterjc:
As of yesterday, we should be able to use the current and default version of mummer 3.23 on BioConda:

https://anaconda.org/bioconda/mummer

I was able to update the recipe to fix the hash-bang issue (which was why #1 pinned to a specific older build of mummer 3.23) and compile it on osx-arm64 as well:

https://github.com/bioconda/bioconda-recipes/pull/49870

PR #44, opened 2024-08-08
Install MiniForge rather than MiniConda

peterjc:
We want to completely avoid the non-free Anaconda defaults channel.

Despite the name, https://github.com/conda-incubator/setup-miniconda is a conda implementation agnostic GitHub Action.

This builds on #42 which dropped the defaults channel.


peterjc:
Looks good, the conda change worked (but still have the new Linux mummer problem noted on #43).

peterjc:
I'll rebase this - should pass now.

widdowquinn (approved):

Issue #13, opened 2024-05-22
Create initial `snakemake` workflow for ANIb 


kiepczi:
To create initial `snakemake` workflow for `ANIb` method, we need to do the following:

1. Update the Makefiles to include the installation of `blast`.

2. Write Initial Workflow to Generate Appropriate Files. 

3. Create Test Set of Target Outputs.

4. Test the Workflow with `pytest`.

peterjc:
Quoting from https://github.com/pyani-plus/pyani-plus/pull/48#issuecomment-2277958288

> My opinion is that we should drop blast-legacy for good with this iteration of pyani. One reason for hanging on to it was to confirm consistency with older ANIb-based tools (e.g. JSpecies), but that doesn't seem so compelling at this point.

peterjc:
I started with just the genome fragmentation on #85, but have moved on to trying to run this within snakemake now...

peterjc:
My work on PR #85 is now part of PR #101, which will close this issue.

I have opened issue #102 for the next step, actually parsing the output generated via snakemake to compute the ANIb identity and log this in the DB.

PR #48, opened 2024-08-09
Simplify thirdparty conda dependencies

### LINKED ISSUE #41: Missing conda packages for dependencies under osx-arm64
Attempting to install the development environment as per the CONTRIBUTING file (cross reference #40), I ran into a problem:

```
PackagesNotFoundError: The following packages are not available from current channels:          
                                                                                                
  - mummer==3.23=h6de7cb9_11
  - gsl==2.7=h93259b0_0
  - fastani
  - blast-legacy
  - blast

Current channels:

  - https://conda.anaconda.org/conda-forge
  - https://conda.anaconda.org/bioconda
```

This was under macOS with an M3 ARM CPU (``Platform: osx-arm64``).

As per https://github.com/bioconda/bioconda-recipes/issues/33333#issuecomment-2221176852 last month, BioConda packages can be updated to cover this platform. I fear mummer and gsl will be problematic given these are pinned to specific (older) builds though. Also BLAST has quite a few dependencies which are not yet all updated.

peterjc:
This does some cleanup of the third-party dependencies:

- both macos and linux files are now sorted for ease of comparison (now identical, if this remains true long term we might merge these)
- dropped currently unused BLAST dependencies (slightly faster for CI; see #13)
- dropped explicit gsl dependency (no longer need to pin specific build on macOS)

With these changes ``conda install --file requirements-thirdparty-macos.txt`` will now run on Apple ARM (``osx-arm64``) and thus almost resolves #41.

### DIFF
--- a/requirements-thirdparty-linux.txt
+++ b/requirements-thirdparty-linux.txt
@@ -1,3 +1,2 @@
 fastani
-gsl=2.7
 mummer=3.23
--- a/requirements-thirdparty-linux.txt
+++ b/requirements-thirdparty-linux.txt
@@ -1,5 +1,3 @@
 gsl=2.7
 fastani
-blast
-blast-legacy
 mummer=3.23
--- a/requirements-thirdparty-linux.txt
+++ b/requirements-thirdparty-linux.txt
@@ -1,3 +1,3 @@
-gsl=2.7
 fastani
+gsl=2.7
 mummer=3.23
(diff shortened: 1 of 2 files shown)


widdowquinn:
My opinion is that we should drop blast-legacy for good with this iteration of `pyani`. One reason for hanging on to it was to confirm consistency with older ANIb-based tools (e.g. JSpecies), but that doesn't seem so compelling at this point.

widdowquinn (approved):

widdowquinn:
That's great - many thanks @peterjc !

PR #54, opened 2024-08-13
Drop unused pytest fixtures

peterjc:
Spotted by hand, confirmed with pytest-unused-fixtures:

```
------- fixtures defined from tests.snakemake.test_snakemake_slurm ------
config_delta_dnadiff_args -- tests/snakemake/test_snakemake_slurm.py:103
config_filter_dnadiff_args -- tests/snakemake/test_snakemake_slurm.py:121
```

### DIFF
--- a/tests/snakemake/test_snakemake_slurm.py
+++ b/tests/snakemake/test_snakemake_slurm.py
@@ -101,42 +101,6 @@ def config_fastani_args(
 
 
 @pytest.fixture()
-def config_delta_dnadiff_args(
-    dnadiff_nucmer_targets_delta_slurm_outdir: Path,
-    input_genomes_small: Path,
-) -> dict:
-    """Return configuration settings for testing snakemake delta rule.
-
-    We take the output directories for the MUMmer delta output and the
-    small set of input genomes as arguments.
-    """
-    return {
-        "outdir": dnadiff_nucmer_targets_delta_slurm_outdir,
-        "indir": str(input_genomes_small),
-        "cores": 12,
-        "mode": "mum",
-    }
-
-
-@pytest.fixture()
-def config_filter_dnadiff_args(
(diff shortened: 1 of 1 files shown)


kiepczi:
Thanks, @peterjc, for spotting this. I think all tests will need to be revised as there might be more fixtures that are either unused or could be improved. This was definitely my oversight, and I will open a new issue for this.

peterjc:
Assuming ``pytest-unused-fixtures`` worked, right now there were not any other unused fixtures defined in pyani-plus.

See https://github.com/mikicz/pytest-unused-fixtures

```
$ pip install pytest-unused-fixtures  # or conda?
$ pytest -v --unused-fixtures  # perhaps with more options
```

kiepczi:
Thanks! I learned something new again!

peterjc:
No problem - I should have put those details in the initial description, sorry.

widdowquinn:
```
$ pip install pytest-unused-fixtures  # or conda?
```

Maybe in `requirements-dev.txt`?

peterjc:
If you think this is worth adding to our standard setup, sure. I'm not familiar enough with the pytest ecosystem to judge - there is also https://github.com/jllorencetti/pytest-deadfixtures which seems similar and also fairly recently updated, but is acknowledged in the ``pytest-unused-fixtures`` README so I assumed that was "newer".

peterjc:
OK. We can install this via ``requirements-dev.txt`` and add ``--unused-fixtures`` to the ``pytest`` commands in the Makefile (local usage) and GitHub Actions workflows (CI usage).

PR to follow.

peterjc:
Cancel that, ``pytest-unused-fixtures`` isn't yet in conda so that makes it non-trivial with our current installation setup.

widdowquinn:
Can we put it under a `requirements-dev-pip.txt` file? I've had to do this in the past when a package isn't under `conda`.

PR #59, opened 2024-08-15
Fix installation of the Python files

### LINKED ISSUE #58: pyani_plus not installed properly, can't import it
## Summary:

Using Intel macOS with the ``pyani-plus_py312`` environment setup as per the CONTRIBUTING file, within the repository root this works:

```
$ python -c "from pyani_plus import __version__; print(__version__)"
0.0.1
```

Elsewhere, it fails:

```
$ cd /tmp
$ python -c "from pyani_plus import __version__; print(__version__)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pyani_plus'
```

Neither an explicit ``pip install -e .`` (almost the same as the Makefile), nor even ``pip uninstall pyani_plus && pip install .`` fixes this.

## Description:

Something would appear to be amiss in the project setup, i.e. ``pyproject.toml``

## Background:

Noticed while working on #51 as when run from a temporary worker directory the snakemake files fail at line 1, ``from pyani_plus.snakemake import snakemake_scheduler``

e.g. From GitHub Actions:

```
ModuleNotFoundError in file /Users/runner/work/pyani-plus/pyani-plus/pyani_plus/workflows/snakemake_fastani.smk, line 1:
No module named 'pyani_plus'
```

Then reproduced locally on Intel macOS.

peterjc:
Closes #58

It seems there is more than one way to express this but this seems explicit and clear:

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ pycodestyle.max-line-length = 120
 ignore = ["ANN101", "ANN102", "COM812", "D203", "D213", "S603"]
 
 [tool.setuptools.packages.find]
-where = ["pyani_plus"]
+include = ["pyani_plus"]
 
 [project.urls]
 Homepage = "https://github.com/pyani-plus/pyani-plus"


widdowquinn:
I see the `src` top directory more frequently these days. Do you (@peterjc @kiepczi) have any thoughts about whether we should adopt it, with `pyani-plus` as a subdirectory? Is there a PEP I've not read/forgotten about?

peterjc:
See https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/ which highlights mostly advantages to using the ``src/`` convention over the more traditional flat layout.

widdowquinn:
> See https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/ which highlights mostly advantages to using the `src/` convention over the more traditional flat layout.

Thanks @peterjc - I see no disadvantages to moving over to `src` layout, other than the work involved.

PR #65, opened 2024-08-19
Helper module of get-path-and-version functions for command line tools

peterjc:
Work towards #32.

Currently based on a mixture of the existing get_version functions in pyani (e.g. pyani/anim.py and pyani/fastani.py) and my similar code in THAPBI PICT: https://github.com/peterjc/thapbi-pict/blob/v1.0.13/thapbi_pict/versions.py

Then tweaked to pass the ruff checks and refactored to reduce duplicated code.

TODO: Do we really need to catch stdout and stderr separately?

### DIFF
--- a/pyani_plus/tools.py
+++ b/pyani_plus/tools.py
@@ -124,14 +124,14 @@ def get_blastn(cmd: str | Path = "blastn") -> ExternalToolData:
     return ExternalToolData(exe_path, version)
 
 
-def get_fastani(cmd: str | Path = "fastani") -> ExternalToolData:
+def get_fastani(cmd: str | Path = "fastANI") -> ExternalToolData:
     """Return FastANI path and version as a named tuple.
 
     We expect the tool to behave as follows:
 
     .. code-block:: bash
 
-        $ fastani -v
+        $ fastANI -v
         version 1.33
 
     Here the function would return the binary path and "1.33" as the version.
--- a/pyani_plus/tools.py
+++ b/pyani_plus/tools.py
@@ -1,0 +1,197 @@
+# (c) The University of Strathclyde 2024-present
+# Author: Peter Cock
+#
+# Contact:
+# peter.cock@strath.ac.uk
+#
+# Peter Cock,
+# Strathclyde Institute of Pharmaceutical and Biomedical Sciences
+# The University of Strathclyde
+# 161 Cathedral Street
+# Glasgow
+# G4 0RE
+# Scotland,
+# UK
+#
+# The MIT License
+#
+# (c) The University of Strathclyde 2024-present
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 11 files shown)


widdowquinn:
IIRC `mummer3` uses an annoying mixture of stderr and stdout, hence that bit of code.

peterjc:
I guess I need to test locally with mummer3 and mummer4, but I'd hope searching stdout+stderr combined would work there too...

widdowquinn:
> I guess I need to test locally with mummer3 and mummer4, but I'd hope searching stdout+stderr combined would work there too...

<clickbait>The results of `nucmer --version > stdout.txt 2> stderr.txt` will surprise you</clickbait>

peterjc:
The codacy analysis seems to be all about subprocess calls, plus using assert in tests which this PR now explicitly ignores for pre-commit.

peterjc:
Huh, ``RuntimeError: fastani not found on $PATH``

I'll review the CI setup as I thought we could assume that was present...

peterjc:
I have no strong view on the use of ``tools.py``, other names could be ``utils.py`` or ``dependencies.py``?

pyani_plus/tools.py
```diff
@@ -0,0 +1,197 @@
...
+    # Turn into an absolute path in case later use a temp working dir:
+    return Path(exe_str).absolute()
+
+
+def _get_path_and_version_output(
+    cmd: str | Path, args: list[str] | None = None
+) -> tuple[Path, str]:
+    """Determine path of command, run it with args, capture combined stdout and stderr."""
+    # Might later need to add check=True as an optional argument,
+    # e.g. NCBI legacy blast doesn't have a version option and uses return code 1.
```
widdowquinn, line 85: I'd like us to drop legacy BLAST with this version of pyani - but as well to keep in mind that not all tools we might use are well-behaved.

pyani_plus/tools.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def get_fastani(cmd: str | Path = "fastani") -> ExternalToolData:
+    """Return FastANI path and version as a named tuple.
+
+    We expect the tool to behave as follows:
+
+    .. code-block:: bash
+
+        $ fastani -v
```
widdowquinn: Should we make this fastANI -v? fastANI should work on both case-insensitive and case-sensitive filesystems.

pyproject.toml
```diff
@@ -23,6 +23,9 @@ select = ["ALL"]
 pycodestyle.max-line-length = 120
 ignore = ["ANN101", "ANN102", "COM812", "D203", "D213", "S603"]
 
+[tool.ruff.lint.per-file-ignores]
+"tests/*" = ["S101"]
```
widdowquinn, line 27: There are a couple more coming in with a merge of the `issue_4` branch. If you wanted to get ahead of that I'm good with adding them in, now.

widdowquinn (approved):

widdowquinn (approved):

widdowquinn:
I'm good with these changes - thanks @peterjc !

pyani_plus/tools.py
peterjc: Yes - I was initially doing this on macOS

PR #67, opened 2024-08-21
Run snakemake workflows with explicit binary paths

### LINKED ISSUE #32: Add module that checks for complete paths to executables

peterjc:
Would close #32, building on #65 and #68.

[Initial draft included work now moved to #68; this branch now continues directly on from that branch via ``git rebase``]

Can now set an explicit path to binary in the snakemake workflow, rather than leaving it as ``nucmer`` and resolved via the ``$PATH`` at run time inside snakemake. This is done via the ``config`` dictionary, re-exported via ``params`` which seems to be a widely used pattern (although I'm not 100% sure why).

Adding configuration to allow the paths to be set explicitly can come later (probably via the optional argument to ``get_nucmer()`` etc). For now this still resolves via the ``$PATH`` but _before_ calling snakemake - and we can then add some logging of the full path and version identified and error handling for missing dependencies.

It seems that I wasn't able to make the first word of the shell command (i.e. the binary name) dynamic with the original ``run:`` ``shell(...)`` syntax, thus the switch to the ``shell:`` syntax.

### DIFF
--- a/pyani_plus/workflows/snakemake_anim.smk
+++ b/pyani_plus/workflows/snakemake_anim.smk
@@ -61,10 +61,8 @@ rule delta:
     input:
         genomeA=get_genomeA,
         genomeB=get_genomeB,
-    run:
-        shell(
-            "nucmer -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
-        )
+    shell:
+        "nucmer -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
 
 
 # The filter rule runs delta-filter wrapper for nucmer
@@ -75,7 +73,5 @@ rule filter:
         "{outdir}/{genomeA}_vs_{genomeB}.delta",
     output:
         "{outdir}/{genomeA}_vs_{genomeB}.filter",
-    run:
-        shell(
-            "delta-filter -1 {input} > {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB}.filter"
-        )
+    shell:
+        "delta-filter -1 {input} > {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB}.filter"
--- a/pyani_plus/workflows/snakemake_anim.smk
+++ b/pyani_plus/workflows/snakemake_anim.smk
@@ -56,11 +56,11 @@ rule delta:
         indir=config["indir"],
         mode=config["mode"],
         outdir=config["outdir"],
-    output:
-        "{outdir}/{genomeA}_vs_{genomeB}.delta",
     input:
         genomeA=get_genomeA,
         genomeB=get_genomeB,
+    output:
+        "{outdir}/{genomeA}_vs_{genomeB}.delta",
     shell:
         "nucmer -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
 
--- a/pyani_plus/workflows/snakemake_anim.smk
+++ b/pyani_plus/workflows/snakemake_anim.smk
@@ -72,6 +72,7 @@ rule delta:
 rule filter:
     params:
         delta_filter=config["delta_filter"],
+        outdir=config["outdir"],
     input:
         "{outdir}/{genomeA}_vs_{genomeB}.delta",
     output:
--- a/pyani_plus/workflows/snakemake_anim.smk
+++ b/pyani_plus/workflows/snakemake_anim.smk
@@ -53,6 +53,7 @@ def get_genomeB(wildcards):
 # the calling code
 rule delta:
     params:
+        nucmer=config["nucmer"],
         indir=config["indir"],
         mode=config["mode"],
         outdir=config["outdir"],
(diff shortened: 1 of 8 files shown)


peterjc:
It looks like mummer v3 ``delta-filter``, ``show-diff``, and ``show-coords`` do not have their own version numbers. If we really need a place-holder value, we'd perhaps just have to assume it matches that of ``nucmer``? I think for now ``get_deltafiler()`` etc will probably return a blank version number (empty string or None).

widdowquinn:
I think those Perl helper scripts from MUMmer could generally be assumed to inherit the nucmer version, but to avoid any edge case confusion we should probably keep it as a null value, as you suggest.

pyani_plus/workflows/snakemake_anim.smk
```diff
@@ -61,10 +61,8 @@ rule delta:
     input:
         genomeA=get_genomeA,
         genomeB=get_genomeB,
-    run:
-        shell(
-            "nucmer -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
-        )
+    shell:
```
widdowquinn, line 65: All seems good, here.

pyani_plus/workflows/snakemake_anim.smk
```diff
@@ -62,7 +63,7 @@ rule delta:
         genomeA=get_genomeA,
         genomeB=get_genomeB,
     shell:
-        "nucmer -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
+        "{params.nucmer} -p {wildcards.outdir}/{wildcards.genomeA}_vs_{wildcards.genomeB} --{params.mode} {input.genomeA} {input.genomeB}"
```
widdowquinn, line 66: Adding this flexibility is a good idea.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ ignore = ["ANN101", "ANN102", "COM812", "D203", "D213", "S603"]
 
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["S101"]
+"tests/generate_fixtures/*" = ["T201"]
```
widdowquinn, line 28: I think it's worth catching leftover `print()` calls that were inserted for debugging purposes. We might want to roll this back unless there's something I'm missing.

tests/generate_fixtures/generate_target_anim_files.py
```diff
@@ -57,11 +59,14 @@
 comparisons = permutations([_.stem for _ in Path(INPUT_DIR).glob("*")], 2)
 inputs = {_.stem: _ for _ in Path(INPUT_DIR).glob("*")}
 
+nucmer = get_nucmer()
+print(f"Using nucmer {nucmer.version} at {nucmer.exe_path}")
```
widdowquinn, line 63: Related to the comment about ignoring T201 - I think it's better to explicitly `# noqa` the odd cases where we _want_ to `print()` than have a bunch of debug `print()`s come through. What are your thoughts?

widdowquinn (changes requested):

tests/generate_fixtures/generate_target_anim_files.py
peterjc: I'll swap over to ``# noqa: T201`` in place of the per-file-ignores. 

I was imagining the fixture generator scripts as being largely self-contained and less likely to ever be hooked into whatever logging framework the main code will use. i.e. Using ``print`` here was and is pragmatic.

widdowquinn: I was thinking that we might use `click` for those fixture generator scripts, which has `click.echo()`. It's not a huge priority, and `print()` is pragmatic, for now.

peterjc:
I've pulled out the new get path/version functions, and the generator script work, into a separate pull request #68.

I'll rebase this WIP onto that new branch - which will just be using the new path functions for calling the workflows.

peterjc:
I'm assuming #68 will be reviewed and merged first - that includes changes following the discussion here about using ``print(...)`` with ``# noqa`` versus a per-file-ignore setting.

peterjc:
One point which puzzled me, but which I think I now understand, is the "config" fixtures in Python have to define all the parameters used in the entire workflow file (including all the new binary paths), even if you are only trying to run one rule which uses just a subset of the config values.

widdowquinn (approved):

widdowquinn:
> One point which puzzled me, but which I think I now understand, is the "config" fixtures in Python have to define all the parameters used in the entire workflow file (including all the new binary paths), even if you are only trying to run one rule which uses just a subset of the config values.

Thanks @peterjc - do you want to briefly document your understanding here, for reference?

peterjc:
My understanding from reading around the topic on Friday is the "config" dictionary is treated as global to each Snakemake workflow file (plus any included files, unless done via modules). It would appear even though we're using the API to run one rule within a workflow file, when the workflow file is parsed Snakemake appears to validate all the config entries have been defined.

So, we can either define all the binaries for all the fixtures (what I did), or perhaps better would be to define the binaries relevant to the rule the fixture is testing and setting the irrelevant binaries to ``/does/not/exist/`` or similar?

widdowquinn:
Thanks @peterjc - that makes sense. In this case I think a valid installation of `MUMmer` should install all of the applications we require, and we'll get away with what's present.

The situations I can picture where a problem might arise seem quite pathological (e.g. the workflow requires binaries A then B, but only A is installed as we're only testing the first stage of the workflow). Always possible I'm missing something, but I'm happy for now with the current solution.

PR #68, opened 2024-08-22
Issue 32 continuation - get path & version functions now used in fixture generation

peterjc:
Continues work towards #32, building on the work merged in #65, by adding functions for the mummer scripts, and now uses all these functions in the fixture generator scripts.

This extracts the non-workflow commits from the WIP on #67, and applies the feedback from there (e.g. using ``print(...)`` with ``# noqa: T201`` in the generator scripts), and gets the test coverage up to full.

This allows us to remove the ``# noqa: S607`` in the generator scripts.


widdowquinn (approved):
All looks good to me.

widdowquinn:
That's great - many thanks @peterjc !

PR #74, opened 2024-08-29
Standardise copyright headers in Python files

peterjc:
I realised some of the Python files were missing a copyright and license block, and there were inconsistencies in the others.

There is scope for a pre-commit hook to enforce this in future.

### DIFF
--- a/pyani_plus/__init__.py
+++ b/pyani_plus/__init__.py
@@ -1,3 +1,39 @@
+# Copyright (c) 2024-present University of Strathclyde
+# Author: Leighton Pritchard
+#
+# Contact:
+# leighton.pritchard@strath.ac.uk
+#
+# Leighton Pritchard,
+# Strathclyde Institute for Pharmacy and Biomedical Sciences,
+# Cathedral Street,
+# Glasgow,
+# G4 0RE
+# Scotland,
+# UK
+#
+# The MIT License
+#
+# Copyright (c) 2024-present University of Strathclyde
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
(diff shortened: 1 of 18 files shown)


peterjc:
I tried https://github.com/johannsdg/pre-commit-license-headers which I liked the look of as it doesn't need yet another configuration file, but the template definition is too rigid to cope with our current usage.

peterjc:
Another tool https://github.com/leoll2/copyright_notice_precommit looks to have no wild card support at all (not even dates), and also no recent changes.

peterjc:
Missing header in ``tests/snakemake/test_scheduler.py``, spotted while testing with ``insert-license`` from https://github.com/Lucas-C/pre-commit-hooks/

peterjc:
This seems to work OK in limited testing:

```
- repo: https://github.com/Lucas-C/pre-commit-hooks
  rev: v1.5.5
  hooks:
    - id: insert-license
      types: [python]
      args:
        - "--license-filepath=LICENSE"
        - "--use-current-year"
        - "--detect-license-in-X-top-lines=20"
        - "--no-extra-eol"
```

Note that this just checks for (and inserts) the MIT license in the Python headers using https://github.com/pyani-plus/pyani-plus/blob/main/LICENSE but that does not cover the individual name and contact information block which our Python files currently start with.

Further tested required, especially in 2025 onward that the current year setting works as expected.

widdowquinn:
Automating this would save a reasonably-sized PITA, and we might like to bend the current licence/copyright info to enable that. To the extent that git blame allows identtification of authors, the traditional "Author:" line is possibly redundant, for instance.

How does `--allow-past-years` work with your testing? It seems like we need both `--allow-past-years` and `--use-current-year` to issue a year range.

widdowquinn:
How does the pre-commit hook manage two copyright statements?

Should we drop the author/address details, retaining only copyright and the (short) MIT licence?

peterjc:
Discussion conclusion: Cut copyright names to just employers (Hutton and/or Strathclyde), drop address, keep full MIT block in text. And work on a pre-commit hook

peterjc:
Pattern matching with more than one copyright holder line is tricky, cross reference https://github.com/Lucas-C/pre-commit-hooks/issues/101

We can solve most of this by using *University of Strathclyde* as the single copyright holder, but that only covers the current team while under contract. It would not handle external contributors, nor any legacy code from the older pyANI project copyright the James Hutton Institute. Still that might suffice in the short term...

PR #77, opened 2024-09-03
Simpler copyright and license headers, with pre-commit check

peterjc:
Follows on from #74 based on our discussion.

This will work for now, but automatically checking matches when there are multiple lines of copyright holders will potentially be a problem. One simple solution is to use something like "Copyright (c) 2024 University of Strathclyde & external contributors" with additional comments above/below the standard block if needed.


peterjc:
Rebased with the new ORM files from #75

widdowquinn (approved):
Turns out simpler is better… ;)

PR #81, opened 2024-09-18
Build MD5 checksum files of FASTA genomes via snakemake

peterjc:
The typical pyANI-plus workflow will be given a folder of FASTA files;
- create new entry in **runs** table
- as needed compute MD5 of each FASTA file, and record in **genomes** table
- as needed compute pairwise ANI over combinations of FASTA files, and record in **comparisons** table

[Ignoring fine details of linking tables etc]

We will be recording the results in a database, but using snakemake to schedule the individual jobs. Currently snakemake v8 is file-centric, and we cannot write rules based on rows in tables.

We can in principle filter the targets before calling snakemake to avoid recomputing where we already have entries in the DB.

Additional commands in the snakemake workflow can potentially parse output files and record their values in the database. The goal of this PR is to explore that for the early stages of our pipeline - computing the MD5 checksum of all the input FASTA files and recording them in the database.


peterjc:
Much of this work ended up going into #83.

As per in-person discussion today, we are unlikely to want to pass the FASTA checksum & parsing for length etc out to worker nodes. We expect this too be quicker performed centrally.

PR #83, opened 2024-09-19
helper functions in db_orm.py for importing data

peterjc:
Based on the https://github.com/widdowquinn/pyani/blob/master/pyani/pyani_orm.py approach with helper functions to record genomes and runs - although this starts simpler with the pre-run tables only.

See also #81 which may be able to use this.

--

Update: See later, this now has helper functions for loading data into all the defined tables, and matching CLI functions too (although we may not need all of the later).

### DIFF
--- a/pyani_plus/private_cli.py
+++ b/pyani_plus/private_cli.py
@@ -48,10 +48,7 @@ def log_genome(
     Any pre-existing duplicate FASTA entries are left as is.
     """
     print(f"Logging to {database}")  # noqa: T201
-    db = Path(database)
-    if not db.is_file():
-        sys.exit(f"ERROR - Database {db} does not exist")
-    session = db_orm.connect_to_db(db)
+    session = db_orm.connect_to_db(database)
 
     file_total = 0
     if fasta:
@@ -101,10 +98,7 @@ def log_comparison(  # noqa: PLR0913
 ) -> int:
     """Log a single pyANI-plus pairwise comparison to the database."""
     print(f"Logging to {database}")  # noqa: T201
-    db = Path(database)
-    if not db.is_file():
-        sys.exit(f"ERROR - Database {db} does not exist")
-    session = db_orm.connect_to_db(db)
+    session = db_orm.connect_to_db(database)
 
     config = db_orm.add_configuration(
         session, method, program, version, fragsize, maxmatch, kmersize, minmatch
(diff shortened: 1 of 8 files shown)


peterjc:
I've moved as far as trying to have multiple worker jobs writing results to the database at once, which was primarily breaking at the genomes table since there would be multiple jobs finishing at similar times using the same genomes - and I had not pre-populated that table.

The GitHub action just now failed with ``MultipleResultsFound: Multiple rows were found when one or none was required`` within ``add_configuration`` when checking if the configuration was already recorded - and the ``session.query()....one_or_none()`` fails. It appears the ``UniqueConstraint`` failed?

peterjc:
As discussed in person today, this PR as it stands is (deliberately) a worst case for causing DB contention. In the test case with 4 genomes, we get all the workers trying to create the DB and empty tables at a similar time point, trying to record the same configuration at the same time point, and then 4 workers trying to record the same genome at about the same time point.

In practice we will create the DB, record the configuration and run, and record all the FASTA file entries in the genomes table on a single thread _before_ invoking snakemake to run the comparisons in parallel.

i.e. The worker nodes will only be recording a comparisons table entry, and while there may be time collisions, they will be trying to write independent rows to the table. A simple pause and re-try should suffice.

peterjc:
A trivial rebase later, this time it failed on macOS Python 3.12, again within ``add_configuration`` with ``session.query(Configuration)....one_or_none()`` raising ``MultipleResultsFound: Multiple rows were found when one or none was required``

peterjc:
Rebased, with upfront creation of the DB and recording of the configuration in the test _before_ calling snakemake. This seems to work on GitHub Actions.

peterjc:
In the current state of the code, there are a bunch of helper functions like ``pyani_plus.db_orn.add_comparison`` for adding rows to the database, with a matching set of CLI functions like ``log_comparison`` for using this from the command line. However, it is likely that we'll only use ``log_comparison`` that way.

There is a full run test in ``test_orm.py`` which records a run before the comparisons, and updates it afterwards. There is also a full test doing all the logging via the CLI functions, but without an easy method to capture the run ID for updating, it simply waits until the end to record the completed run in one go.

pyani_plus/private_cli.py
```diff
@@ -0,0 +1,379 @@
...
+        sim_errors=sim_errors,
+        cov_query=cov_query,
+        cov_subject=cov_subject,
+    )
+
+    session.commit()
+    return 0
+
+
+def parse_fastani_file(filename: Path) -> tuple[Path, Path, float, int, int]:
```
peterjc, line 257: We can move ``parse_fastani_file`` (and perhaps ``log_fastani``) into a methods module later (I didn't want to conflict with the ongoing ANIm and dnadiff work).

Issue #63, opened 2024-08-19
Consider reducing test dataset sizes


widdowquinn:
## Summary:

In the old `pyani` we used _Blochmannia_ genome example datasets (among other things), and moving to `pyani-plus` gives us an opportunity to update our choices to reduce test run time and target particular known (prior) issues.

For example, we could introduce Donovan Parks' virus sequences (https://github.com/widdowquinn/pyani/issues/340), which would allow us to simultaneously reduce input test dataset size and confirm that our fixes for the coverage problem continue to work (and validate ANIm/other tests).

widdowquinn:
We may want to consider dividing tests of datasets into groups (e.g. "fast" and "slow"; "viral", "bacterial", "linear genomes") etc. and running these on different schedules. For example:

- "fast" tests would run in CI on a `push`
- "slow" tests would run in CI on a PR merge
- "complete" tests might run monthly on CI/locally when being particularly paranoid about a change.

peterjc:
See e.g. https://docs.pytest.org/en/7.1.x/example/markers.html#registering-markers which uses marking some tests as "slow" for example.

widdowquinn:
This would help avoid the current duplicate `push` and `pull-request` GH Action runs we have at the moment. Running the "fast" tests only on `push` and the "slow"/"complete" tests on `pull-request` would be a useful, meaningful separation of the two activities.

PR #85, opened 2024-09-25
Add expected ANIb fragment and BLAST files to test fixtures

peterjc:
This is a step towards #13, adding ANIb support.

Initially I copied the four small genome fragment files from the old pyani test suite, they are a little large though (41MB in all):

https://github.com/widdowquinn/pyani/tree/master/tests/fixtures/anib/fragfiles

Instead, after verifying new script ``generate_anib_fragment_files.py`` reproduced those files, I have switched to the tiny two genome viral example outlined in #63 (Donovan Parks' virus sequences). See #87, which this branch is currently based on.

Update: Now rebased onto main following #98 which shifted to just three similar viruses for our main test case. I have the next steps of building the BLAST databases and running blastn working, need to finish the "make fixtures" support first.

Update: Now covers the FASTA fragment files (queries) and BLAST databases (subject) via snakemake with tests and generate fixtures support. Final blastn step working locally, need to write the fixture generator.


peterjc:
I'll split this WIP up, moving the switch to the two viral genomes for the snakemake tests to a separate PR.

peterjc:
Closed in favour of #101 which extended this enough to close the issue.

PR #61, opened 2024-08-16
Drop duplicate tests

peterjc:
See discussion on #34. Reverts #50.

I've looked at the fastani test, and satisfied myself it was a near duplicate (only changes to the output path, and the number of cores).

PR #86, opened 2024-09-25
Some minor cleanup

peterjc:
See commit comments, some of this is things I missed in #61.


peterjc:
I repeated the ``pytest -v --unused-fixtures`` check, looks clear now. Cross reference #54

PR #87, opened 2024-09-25
Adopt two viral genomes for the slow snakemake tests

### LINKED ISSUE #63: Consider reducing test dataset sizes

peterjc:
This would close #63 by switching the anim and dnadiff snakemake tests from the current ``input_genomes_small`` example (4 bacteria) to ``input_genomes_tiny`` (just 2 viruses). It would leave the fastANI snakemake test as is (using the 4 bacteria).

Note currently these tests do not check the actual ANI values (yet), so the change is mainly regenerating the fixtures. See also minor cleanup preceding this on #86.

Locally this brings the test run time down to 15s (Intel macOS).


peterjc:
The GitHub actions now run the tests themselves in under 10s, leaving the CI setup as the major overhead.

Miniforge alone looks like it take about 30s on Linux, 1m30s on macOS - and then about the same again for our dependencies. We may be able to speed that up with better caching...

widdowquinn:
Can we add a third/fourth virus to have (yet another) reassurance that multi-way pairwise comparisons are performed?

widdowquinn:
As discussed - the current release of `pyani` should be a good ground truth (after @kiepczi fixed the issue raised by Donovan) for the viral comparison results.

peterjc:
Rebased after #89.

peterjc:
@kiepczi note since this drops the old ANIm targets, for issue #4 etc you may need to restore selected files and/or switch to using the viral examples?

kiepczi:
> @kiepczi note since this drops the old ANIm targets, for issue #4 etc you may need to restore selected files and/or switch to using the viral examples?

I’m currently working on this, but we might need a different set of viral genomes. These genomes are too diverse, and they return a `ZeroDivisionError` in `pyANI v0.3`, so I can’t use them for testing aniM values.

As we discussed yesterday, we want to ensure that in `pyani_plus`, the `ZeroDivisionError` is handled, and for comparisons where there is no alignment, we will return `NULL`.

I’ll look for a different set of genomes to test the aniM calculations and multi-way pairwise comparisons. I’ll also keep the genomes that don’t align, we definitely need them for other things.

PR #90, opened 2024-09-28
Adding tools.get_makeblastdb helper function

peterjc:
Will need this for #13


widdowquinn:
I'm ambivalent about recording `makeblastdb` version/parameters in the database. I don't think it hurts but - as discussed - it adds a wee bit of complexity (maybe an additional field for "other programs" might be useful, populated with a JSON entry?). We might revisit that.

widdowquinn (approved):

PR #93, opened 2024-10-01
Improve test coverage for DB logging and fastANI logging

peterjc:
I think to get the ORM coverage back to 100%, we may need a test with worker threads deliberately writing to the DB at the same time. This is perhaps worth trying for reassurance that this will work for real.


peterjc:
We can do better, these lines in ``private_cli.py`` should be easy to cover from ``tests/test_private_cli.py``:

```python
        msg = f"ERROR: Database {database} does not exist, but not using --create-db"
        sys.exit(msg)
```

peterjc:
According to codecov.io, stuck at ``db_orm.py`` 94.59% and ``private_cli.py`` at 94.50% but I think those numbers are stale...

peterjc:
According to codecov.io, now at db_orm.py 95.14% and private_cli.py at 94.50%.

This could trigger the addition of a genome via another thread while the MD5 was computed - but no further down the exception handling in the add-genome function (which was originally being stress-tested deliberately).

peterjc:
The remaining untested lines are corner cases where something has gone wrong, in some cases perhaps conceptually better written as an assert since I don't really expected it to break (but then of course the ruff check would object).

peterjc:
I'll rebase this for the small/tiny example change...

tests/test_private_cli.py
```diff
@@ -251,6 +251,22 @@ def test_log_comparison_parallel(tmp_path: str, input_genomes_small: Path) -> No
     )
 
     fasta = list(input_genomes_small.glob("*.fna"))  # subset of folder (2)
+    pool = Pool(len(fasta))
```
peterjc, line 254: We changed the extensions so there is now only one *.fna, so should probably hard code the pool size larger.

widdowquinn:
As ever, predicting the behaviour of future parallel runs is tricky - are you sure we can make this fail 100% deterministically? If not, then we might not want to spend _too_ much time on it.

peterjc:
Will need rebasing for the change from 4 bacteria to 3 viruses for the fastANI (and other) tests...

PR #95, opened 2024-10-02
New set of viral genomes for testing

### LINKED ISSUE #94: Need New Viral Genome Set for ANI Testing
We talked about using the values reported by `pyani v3` as the ground truth after fixing the genome coverage issue. However, the current set of viral genomes (comprising 4 genomes) is too diverse, resulting in a `ZeroDivisionError`. 

To ensure that multi-way comparisons are generated and that we can validate the ANI values against those returned by `pyani v3`, we need a set of viral genomes consisting of at least four genomes, with at least a portion of each genome aligning.

Additionally, it would be beneficial to obtain another set of genomes that will not align. This will help us test the handling of the `ZeroDivisionError` to ensure it correctly returns zero when applicable.

kiepczi:
This PR fixes #94. 

We will now have 3 sets of genomes for testing: 
**Set 1: viral_example** four viral genomes, for which all comparisons return meaningful alignments. 
**Set 2: bacterial_example** for bacterial genomes, needed to test `fastani` option
**Set 3: failure_example** four viral genomes, for which some comparisons return empty alignments. This is needed to handle `ZeroDivisionError`.


peterjc:
There seem to be only two viral genomes in the new example (not four), what are ``genome_3.fasta`` and ``genome_4.fasta`` and do they need better names?

kiepczi:
My apologies the naming of the genomes might not be the best, and I'm happy to change this to something else. There are, however, 4 genomes:
- MGV-GENOME-0264574.fna
- MGV-GENOME-0266457.fna
- genome_3.fasta
- genome_4.fasta

I wanted to keep the MGV-GENOME-0264574.fna and MGV-GENOME-0266457.fna as these are the "problematic" genomes provided by Donovan a while ago. The other two genomes (genome_3 and genome_4), I downloaded from NCBI after finding these were very similar to the MGV-GENOMOES.

peterjc:
It may be GitHub misleading me then - it looked like just 3 and 4 in the directory.

peterjc:
Do we need/want the bacterial example anymore? We're currently still using it for fastANI, but only because it is fast enough (otherwise I'd have switched it to the viruses). Probably a Leighton question...

kiepczi:
> Do we need/want the bacterial example anymore? 

You are right. We might not need them any more. 
I guess it won't reduce time for testing substantially, but it will make `tests` directory a wee bit neater, or we could rearrange things a little bit if we decide to keep them. Either way let me know, what you prefer and I will take care of things ;)

peterjc:
Thinking about this, right now we don't need multiple sets of genomes at all - we just need one set for the snakemake tests where we also have the intermediate files under the fixtures. And this should be as small and fast as reasonably possible (so maybe 3 rather than 4 viruses).

Is there anything stopping you using the current virus set for testing ANIm calculation? If so, maybe we should just revert b7522253954e3a8e0a18ae472b619f8053762301 and b2b1c227350c3131653818c249935c86f18f49ac and instead add one of your similar viruses.

Separately, shall I switch the fastANI snakemake tests to use the same virus set (meaning the bacterial set will be unused in the short term)?

Later we will want additional sets of genomes for testing things like reports and plotting - but there I see no reason to record the intermediate files in git for testing.

kiepczi:
There are two options here. 
**Option 1**
One set containing four viral genomes, where all comparisons return an alignment. This would help verify that everything is working as expected, including multiway comparisons, calculations, etc. This set is currently in `viral_example`.
Another set of genomes where not all comparisons return an alignment. This would allow us to ensure that `ZeroDivisionError` is handled correctly. Treating these tests separately would give us a clear distinction when things go wrong. This set is currently in `failure_example`.
**Option 2**
Alternatively, we could revert to one of the previous commits, as you suggested, and keep a single test set with four genomes, where two genomes do not return an alignment.

My only issue with option 2 is that I will not be able to run the older version of pyANI on all four genomes to obtain matrices to verify the accuracy of the calculations. This is because pyANI fails to complete the comparisons due to a `ZeroDivisionError`. Yes, we can run `pyANI v0.3` on two genomes, which will rerun an alignment to obtain matrices for other testing purposes. 

One thing that might be worth remembering is that there may be a slight issue with the calculations for `dnadiff` method. I have run into problems before where my calculated `dnadiff` values occasionally differ from those returned by `dnadiff` software, because the content of the `.rdiff` files can sometimes vary between runs. The tool identifies gaps in the alignment using graphs, I believe, which can sometimes result in different values.

peterjc:
Ah - I was under the impression pyANI v0.3 could handle the non-aligning cases, that does make dealing with badly aligning genomes a little harder to deal with long term as we'd have to compile the expected results "by hand".

kiepczi:
> Ah - I was under the impression pyANI v0.3 could handle the non-aligning cases. 

Unfortunately it does not. Should we go with option 1, and have two different sets for both cases for "working" and "failing" genomes?

peterjc:
Option three, a small set of three viruses which all give alignments - #98 ?

And later test non-alignments explicitly which doesn't have to be part of an all-vs-all example.

peterjc:
Handled via https://github.com/pyani-plus/pyani-plus/pull/98

PR #98, opened 2024-10-03
Adjust the virual test case

peterjc:
Alternative to #95, avoid the non-aligning viruses and instead add a third closer sequence.

I would propose to go further and switch the fastANI test from the current bacteria to this virus test set too.


tests/fixtures/anim/targets/delta/MGV-GENOME-0264574_vs_OP073605.delta
```diff
@@ -1,2 +1,8 @@
-/Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0264574.fna /Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0357962.fna
+/Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0264574.fna /Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/OP073605.fasta
 NUCMER
+>MGV_MGV-GENOME-0264574 OP073605.1 39253 57793
+107 39253 57793 18644 27 27 0
+-2991
+-24624
+-1
+0
```
peterjc, line 8: GitHub is showing the overall diff for the pull request as renaming and editing this file.

As per the history, we delete the old file (which had no alignment), and add a new files (which does have an alignment).

peterjc:
CodeFactor is complaining about the MD5 checksum as usual - since I touched a file using this.

kiepczi:
I like this alternative to #94. I think we could take care of two more things before merging:

- Remove bacterial genomes and use the viral examples for FastANI testing.
- Change one more file extension to `.fas` so that we test all the possible extensions we want Pyani-plus to accept (e.g., 'fasta', 'fna', and 'fas')

peterjc:
I'd change the fastANI bacteria example separately - we both have lots waiting on this.

Note with the extension change, it makes ``test_log_comparison`` a little less realistic as it now only logs a single self-vs-self comparison. But that is still enough for that minimal test.

peterjc:
Do you think we should delete the bacterial genomes? Or leave them there unused in the short term?

kiepczi:
> I'd change the fastANI bacteria example separately - we both have lots waiting on this.

We can do this separately. Not a big deal I guess. 

> Note with the extension change, it makes `test_log_comparison` a little less realistic as it now only logs a single self-vs-self comparison. But that is still enough for that minimal test.

I didn’t realise this—my bad. I also think it's important to ensure that all possible FASTA extensions are taken into consideration.

PR #96, opened 2024-10-02
parse aniM output and calculate aniM values

kiepczi:
The code in this PR was developed to calculate aniM values. It might be a good idea to create another issue to handle adding these values to the database, since that hasn't been done yet.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [X] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [X] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [X] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [X] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [X] Create a new branch with a short, descriptive name
- [X] Work on this branch
  - [X] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [X] new code is commented, especially in hard-to-understand areas
  - [X] corresponding changes to documentation have been made
  - [X] tests for the change have been added that demonstrate the fix or feature works
- [X] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [X] Rebase against `origin/master`
- [X] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [X] Commit branch
- [X] Submit pull request, describing the content and intent of the pull request
- [X] Request a code review
- [X] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #100, opened 2024-10-04
add fixtures for anim method (expected results)

kiepczi:
As suggested by @peterjc in #96, the history of work done on the `issue_4` branch is complex and might be difficult to follow. Splitting this into smaller branches and merging them in stages would be a better option.

In this PR, I provide:

- A single script to generate matrices using `pyaANI v0.3`, which will be required for testing the aniM method calculations.
- Generated fixtures: `pyANI` matrices for a set of three viral genomes.
- An updated `conftest.py` fixture to point to the directory containing the matrices.


tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,48 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
```
peterjc, line 24: I would suggest adding the line ``set -euo pipefail`` here for strict bash mode, see for more http://redsymbol.net/articles/unofficial-bash-strict-mode/ but the key bit is ``set -e`` means bash will abort on the first error and not keep going.

peterjc: (I am torn on having this as line 2 above the license, or immediately after the license header)

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,48 @@
...
+#create db
+pyani createdb --force
+
+#index with pyANI v0.3
+pyani index -i ../../tests/fixtures/viral_example/
+
+#Update the label file so that the columns in the matrices correspond to the genome stem files
+awk -F'\t' 'OFS="\t" {$3 = $2; print}' ../../tests/fixtures/viral_example/labels.txt > ../../tests/fixtures/viral_example/labels_tmp.txt && mv ../../tests/fixtures/viral_example/labels_tmp.txt ../../tests/fixtures/viral_example/labels.txt
```
peterjc, line 32: Can we do something clever here to insert the MD5 checksums as the labels?

Perhaps using the output of ``md5sum ../../tests/fixtures/viral_example/*.f*``?

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,48 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+#create db
+pyani createdb --force
+
+#index with pyANI v0.3
+pyani index -i ../../tests/fixtures/viral_example/
```
peterjc, line 29: I would suggest making symlinks to the FASTA files in the temp working folder, that way there are no stray ``*.md5`` files etc to clean up.

On linux this would be just ``cp -s ../../tests/fixtures/viral_example/*.f* .`` but on macOS you'd have to use ``ln`` in a loop:

```bash
for FASTA in ../../fixtures/viral_example/*.f*; do ln -s $FASTA ${FASTA##*/} ; done
```

(This would need adjustment if you want to put the FASTA links in a subfolder - ask)

I assume pyANI is also fine with ``.fasta``, ``.fas``, and ``.fna`` here?

tests/fixtures/anim/matrices/runs.tsv
```diff
@@ -0,0 +1,2 @@
+	run ID	name	method	date run	command-line
+0	1	genearte fixtures	ANIm	2024-10-04 11:17:26.020690	/opt/anaconda3/envs/pyani_issue_421/bin/pyani anim -i ../../tests/fixtures/viral_example/ -o ../../tests/fixtures/viral_example/output -v -l ../../tests/fixtures/viral_example/output/log.log --name genearte fixtures --labels ../../tests/fixtures/viral_example/labels.txt --classes ../../tests/fixtures/viral_example/classes.txt
```
peterjc, line 2: I don't think we need/want the ``runs.tsv`` file

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
peterjc, line 1: I would use as ``#!/bin/bash`` per http://redsymbol.net/articles/unofficial-bash-strict-mode/ but https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang has some good discussion as there are pros-and-cons. This works on macOS and most if not all Linux.

peterjc (changes requested):
See comments, e.g. remove ``runs.tsv`` as not needed.

tests/generate_fixtures/generate_pyani_anim_matrices.sh
kiepczi: Changing `{$3 = $2; print}` to `{$3 = $1; print}` fixed this issue. The first column in `labels.txt` already provides md5sum :)

tests/generate_fixtures/generate_pyani_anim_matrices.sh
kiepczi: The `ln` loop worked fine :)

Yes, `pyANI v0.3` works on these extensions so there is no problem with that at all.

tests/fixtures/anim/matrices/runs.tsv
kiepczi: Fair. removing `--runs` parameter from `pyani report` cmd prevents from generating the `runs.tsv` file. I have now removed the file from the repo.

kiepczi:
Thanks, @peterjc, for the review. I have now made the requested changes. Once this is all approved, I will create a PR with the aniM calculations. I will need the fixtures to test the code.

In the meantime, I will do the same for the dnadiff method. The `issue_17` branch is probably as messy as `issue_4`.

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,58 @@
...
+#Run comparisions
+pyani anim -i . -o output -v -l output/log.log --name "genearte fixtures" --labels labels.txt --classes classes.txt
+
+#Generate matrices
+pyani report -v -o ../../../tests/fixtures/anim/matrices/ --formats=stdout --run_matrices 1
+
+#Change extensions files
+for file in ../../../tests/fixtures/anim/matrices/*; do
+    mv "$file" "${file%.tab}.tsv"
+done
```
peterjc, line 54: Rather than generating the files there, and then renaming them in place, I would find it clearer to generate them locally, and then move and rename to the output fixtures directory?

Also, ``for file in .../*.tab; do`` with the ``.tab`` explicit seems safer.

tests/generate_fixtures/generate_pyani_anim_matrices.sh
peterjc: Easier than I expected. How does this look with the new output?

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,58 @@
...
+
+#index with pyANI v0.3
+pyani index -i .
+
+#Update the label file so that the columns in the matrices correspond to the genome md5 hash
+awk -F'\t' 'OFS="\t" {$3 = $1; print}' labels.txt > labels_tmp.txt && mv labels_tmp.txt labels.txt
+
+#Run comparisions
+pyani anim -i . -o output -v -l output/log.log --name "genearte fixtures" --labels labels.txt --classes classes.txt
```
peterjc, line 46: Typo, genearte -> generate

tests/generate_fixtures/generate_pyani_anim_matrices.sh
```diff
@@ -0,0 +1,58 @@
...
+# Make a temp subdir
+mkdir tmp_pyani
+cd tmp_pyani
+
+# Create pyanidb
+pyani createdb
+
+# Copy genomes to tmp_pyani subdir (macOS)
+for FASTA in ../../../tests/fixtures/viral_example/*.f*; do
+    ln -s $FASTA ${FASTA##*/};
```
peterjc, line 36: I just put the whole script through https://www.shellcheck.net/ (they have an online checker, I am wondering about adding this to our pre-commit setup - see #103), and it only had one issue to flag:

```
    ln -s "$FASTA" "${FASTA##*/}";
```

i.e. Double quote the filenames on the off chance they contain spaces.

peterjc:
The new script needs to be added to the ``Makefile`` too, I suggest right after ``generate_target_anim_files.py``

peterjc (approved):
I assumed you were in the 3pm meeting, so went ahead with the other change suggestions.

peterjc:
I could expand this to do fastANI as well (and will start a new branch), but not ANIb for testing #102. As noted there, it seems the pyANI v0.3 master branch for ``subcmd_anib.py`` and ``subcmd_aniblastall.py`` are full of ``NotImplementedError`` stubs. Is there an unfinished branch I should look at perhaps?

kiepczi:
> I assumed you were in the 3pm meeting, so went ahead with the other change suggestions.

Thank you! I indeed went to 3pm meeting. Is this ready to be merged, so I can start working on calculating aniM values?

kiepczi:
> I could expand this to do fastANI as well (and will start a new branch), but not ANIb for testing #102. As noted there, it seems the pyANI v0.3 master branch for `subcmd_anib.py` and `subcmd_aniblastall.py` are full of `NotImplementedError` stubs. Is there an unfinished branch I should look at perhaps?

Maybe @widdowquinn can help with this. I haven't work on ANIb method, and from my understanding everything should be working fine.

peterjc:
I think this PR is ready to go. Personally I'd click the "rebase and merge" green button.

PR #103, opened 2024-10-04
shfmt & shellcheck in pre-commit setup for consistent shell script formatting

peterjc:
We are starting to get more shell scripts in pyANI plus, currently in the tests (e.g. mock binaries for testing version detection, and for regenerating fixtures). As with the Python and snakemake files, a consistent style would be nice - such as the tool ``shfmt`` https://github.com/mvdan/sh#shfmt

I have been using this on https://github.com/peterjc/thapbi-pict/blob/master/.pre-commit-config.yaml


peterjc:
Once the current batch of pending pull requests is done, and we have a few more shell scripts on the main branch, it may be worth adding ``shellcheck`` too - https://www.shellcheck.net/

Again I have been using this before, but it was quite strict. On the plus side, I learnt a lot more about bash through resolving its complaints.

peterjc:
The fixture generating script on #100 passes with ``shellcheck`` now, so adding that here too.

PR #105, opened 2024-10-04
Compare fastANI matrix output to that from pyANI v0.3 - Issue 104

peterjc:
This is still a WIP, but covers the basics - just needs some additional work to get the test to pass. To start with, currently rather than 100% identity have NaN for self-self entries.


peterjc:
Ah, we currently don't include the self comparisons in the targets fixtures:

```python
# Running comparisons
comparisons = permutations([_.stem for _ in Path(INPUT_DIR).glob("*")], 2)
```

Note:

```pycon
>>> list(permutations(["a", "b", "c"], 2))
[('a', 'b'), ('a', 'c'), ('b', 'a'), ('b', 'c'), ('c', 'a'), ('c', 'b')]
```

peterjc:
I think this is probably ready to merge now even though it will need followup work to resolve why the other matrices don't match (currently just the identity matches), but this gives us a framework to adopt for the ANIm etc tests too.

peterjc:
We discussed all-vs-all and the need for explicit self-vs-self comparisons. We might assume 100% identity and coverage for self-vs-self (as assumed in pyANI v0.3), but my worry was algorithmic details might break this (especially coverage due to filters dropping regions).

Leighton's quick test (three copies of the same 4Mbp bacterial genome) showed no change on ANIm, but coverage on ANIb did drop to 99.999% (lost about 30bp). There _should_ be no change on fastANI or sourmash style approaches. This issue is likely most obvious on ANIb.

Doing self-vs-self is therefore probably wise on all the methods (and we can then check if a shortcut can be justified based on real examples).

widdowquinn:
I'd be inclined to add self-vs-self to all comparisons and report departure from theory (i.e. 100% identity/100% coverage) in its own table. As noted, I expect ANIb to be the only method departing from this because of internal heuristics (though who knows with MUMmer…)

peterjc:
From fastANI output line of (query name, ref name, ani, number of orthologous matches, number of query fragments) we want to use the ratio (number of matches / number of fragments) as our query coverage (a proxy for how the other methods work). Can _perhaps_ stretch this to estimate the alignment length... and make an argument for sim-errors.

widdowquinn:
Not such a fan of the alignment length being retained myself, but I see the analogy. Sim errors (count of mismatching fragments) makes more sense. So long as we're clear in the docs what these measures represent, method-by-method, there's no harm in including them.

peterjc:
Noting that like the ANIb algorithm, fastANI also does not always report 100% ANI for self-vs-self, see https://github.com/ParBLiSS/FastANI/issues/132

PR #107, opened 2024-10-08
Issue 4 aniM calculations

kiepczi:
As agreed in #96, work done on `issue_4` needed to be split into smaller chunks. 

In this PR, I propose the following:

- Adding `intervaltree` to `requirements.txt`. While working on the legacy `pyANI`, we settled on using this package to merge overlapping regions that are still present in `.deltafilter` files.
- `method_anim.py` is a script used to calculate aniM values, which was tested and found to be congruent with the legacy `pyANI` in `tests/methods/test_anim.py`.


pyani_plus/methods/method_anim.py
```diff
@@ -0,0 +1,271 @@
...
+
+    The passed directory should contain the output of nucmer comparisons for a single
+    run on a set of input sequences. Each comparison is stored in a separate file,
+    parsed to obtain the count of aligned bases from the query and reference
+    sequences, similarity errors, and average nucleotide identity for each comparison.
+
+    :param deltafilter: Path to the filter file
+    :param indir: Path to input directory (FASTA files)
+    """
+    genome_lengths = {
```
kiepczi, line 241: This might not be needed if we decide to use `md5` hashes in the file names of comparisons, and not filenames (eg. `<genome_1_md5>_vs_<genome_2_md5>.filter` rather `<genome_1_vs_genome_2>.filter`).

peterjc:
I would be inclined to use ``tests/test_anim.py`` rather than ``tests/methods/test_anim.py`` (and did this for ``tests/test_fastani.py``).

pyani_plus/methods/method_anim.py
```diff
@@ -0,0 +1,254 @@
...
+    r_length: int  # total base count of reference sequence
+    q_cov: float  # query coverage (as a percentage)
+    r_cov: float  # reference coverage (as a percentage)
+    r_hadamard: float  # reference hadamard (percentage identity * percentage coverage)
+    q_hadamard: float  # query hadamard (percentage identity * percentage coverage)
+
+
+# NOTE This might not be needed as this information can now
+# be stored in one of the dataframeas in the database.
+def get_genome_length(filename: Path) -> int:
```
peterjc, line 80: You shouldn't need this but it will need some rearranging to have access to the DB at the time the value is used.

pyani_plus/methods/method_anim.py
```diff
@@ -0,0 +1,254 @@
...
+    return aligned_bases
+
+
+def parse_delta(filename: Path) -> tuple[int, int, float, int]:
+    """Return (reference alignment length, query alignment length, average identity, similarity errors).
+
+    :param filename: Path to the input .delta file
+
+    Calculates the aligned lengths for reference and query and average nucleotide
+    identity, and returns the cumulative total for each as a tuple.
```
peterjc, line 114: Also returns similarity errors.

kiepczi:
Thanks @peterjc for the code review. 
I have now moved the `test_anim.py` from `tests/methods` to `tests` (4481a93), and updated the `parse_delta` doctoring (d8c680f).

pyani_plus/methods/method_anim.py
kiepczi: Do you think that should be addressed under this branch now, or perhaps under a new one?

peterjc: Probably a new PR - I suspect it will need some refactoring.

tests/test_anim.py
```diff
@@ -0,0 +1,219 @@
...
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
```
peterjc, line 42: Another doubled license block - perhaps the pre-commit hook needs tweaking (I think there is a fuzzy match option)

peterjc (changes requested):
I think once the double-license is fixed, we can merge this 👍

peterjc (approved):

kiepczi:
> I think once the double-license is fixed, we can merge this 👍

Great! Merging now :)

PR #108, opened 2024-10-08
Expected ANIb matrices computed with pyANI v0.2 - Issue 102

peterjc:
This is the first step to issue #102, still need to replicate the calculation of ANIb and then cross-check the resulting matrices against these.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [x] corresponding changes to documentation have been made
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [ ] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [ ] Request a code review
- [ ] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

Issue #102, opened 2024-10-04
Port ANIb code from legacy pyANI to compute and log identity etc


peterjc:
This will build on #13 which handles the query FASTA file fragmentation, subject file BLAST database building, and actually running NCBI BLAST to compare the fragmented query to the subject database.

This issues is to:

* Parse the tabular output from ``blastn`` to compute the identity etc as per ANIb
* Use that to define another private-cli function to log an ANIb comparison (see the fastANI code added initially in 7b524ed5921612f01279cf26ea4052d926ef08d9 for this approach)
* Extend the snakemake workflow test to setup the DB and then call this new functionality (see the fastANI test code https://github.com/pyani-plus/pyani-plus/blob/main/tests/snakemake/test_snakemake_fastani_workflow.py for this)

peterjc:
Related to this is generating reference values to test against, but it seems the pyANI v0.3 master branch for ``subcmd_anib.py`` and ``subcmd_aniblastall.py`` are full of ``NotImplementedError`` stubs.

widdowquinn:
Yes - sorry about that. It was a thing to be done earlier in the project but that wasn't reached :(

Example code is in the `version_0_2` branch of old `pyani', e.g. https://github.com/widdowquinn/pyani/blob/version_0_2/pyani/anib.py

peterjc:
Success with something based on ``average_nucleotide_identity.py -m ANIb -i . -o output -v -l ANIb.log --force`` giving ``ANIb_percentage_identity.tab`` etc:

```
                    OP073605            MGV-GENOME-0266457  MGV-GENOME-0264574
OP073605            1.0                 0.9946458974358975  0.9992984615384616
MGV-GENOME-0266457  0.99467             1.0                 0.9944781578947369
MGV-GENOME-0264574  0.9993184615384614  0.9945938461538463  1.0
```

Note to self: pyANI v0.2 had a different main script name than v0.3

PR #109, opened 2024-10-08
Discussion - ANIm files using pyani v0.2 vs 0.3 differ

peterjc:
For issue #102 (groundtruth ANIb using pyani v0.2, see PR #108) we need to use pyANI v0.2 rather than v0.3 when we generate the expected matrices. I thought it would be nicer therefore to also use v0.2 for generating the ANIm groundtruth values - assuming they would be the same.

The script to do this for ANIm was essentially a copy of the proposed ANIb script with a search-and-replace.

This PR is to demonstrate changes in the ANIm output from v0.2 (specifically the [version_0_2](https://github.com/widdowquinn/pyani/tree/version_0_2) branch as of [4c19123c67b78c01d58869527402aec518ead473](https://github.com/widdowquinn/pyani/commit/4c19123c67b78c01d58869527402aec518ead473) to v0.3 (specifically the main branch as of [6b8b262b028f7d27c2dfdddbeda67e7cb571fee7](https://github.com/widdowquinn/pyani/commit/6b8b262b028f7d27c2dfdddbeda67e7cb571fee7)).

First there is a trivial difference in the row/col labels switching order from v0.2 to v0.3. Therefore this PR explicitly sorts them matrix by the MD5 (as being done anyway in this repo internally as part of testing against the expected matrices).

The final commit dc87d7eb2287574ec743bf25d423d292612fa05e on this PR shows meaningful differences including:

* Alignment lengths were floats (ending ``.0``) under v0.2, and included some longer than the shorter sequence (!)
* Some coverage values of approx 1.5 (see above)
* Knock on changes to the Hadamard
* Sim-errors diagonals were floats (ending ``.0``) under v0.2 with a zero diagonal, but have got diagonal values of ``1`` under v0.3 (!)

i.e. 

* This looks like a bug in v0.2 (overly long alignments) which was fixed in v0.3.
* There looks to be a quirk in v0.3 with non-zero self-vs-self values of one for sim-errors


peterjc:
(This test will pass right now as we've not yet merged #107 which will check the pyANI plus ANIm calculations against these expected values)

peterjc:
Confirmed with @kiepczi this is due to the bug in v0.2 fixed in v0.3 by https://github.com/widdowquinn/pyani/pull/425

This means we will need to split the makefile fixtures entry as we need both pyani 0.2 (for ANIb) and 0.3 (for ANIm).

See also https://github.com/pyani-plus/pyani-plus/commit/7e41764d8573d27e4e9ade4e964b01bbebc741e6 which I made to stop accidentally using the wrong pyani - similar guard in #108.

peterjc:
Closing in favour of issue https://github.com/widdowquinn/pyani/issues/438

Issue #17, opened 2024-06-05
Write module to parse `dnadiff` output


kiepczi:
After developing the initial workflow for the `dnadiff` method #12, we need to ensure that we can reproduce the `AlignedBases` value provided by `dnadiff`.

To achieve this, we should:
- Write a module to parse all files generated by the Snakemake workflows for the `dnadiff` method.
- Generate a test set of expected output using `dnadiff`.
- Use `pytest` to compare the expected output with the output provided by the module.

kiepczi:
I started working on this under the `issue_17` branch. 

So far, I have written functions to parse `.mcoords` and `.rdiff` files. The code still requires testing and few tweaks.

kiepczi:
Should we close this issue since parsing of `dnadiff` output was merged with #126?

peterjc:
Yes, and #133 will cover the final steps of getting this into the DB.

PR #110, opened 2024-10-08
add fixtures for dnadiff method (expected results)

kiepczi:
As stated in #17, we need to write a module that will parse the `dnadiff` output. I have been working on this in the `issue_17` branch, but the `main` branch has changed _slightly_ since then, and the working history on this matter might be hard to follow.

Similar to #96, I have decided to break this into smaller chunks and merge smaller changes into the `main` branch.

This PR introduces the following:

- Minor changes to `generate_target_dnadiff_files.py` to generate `.report` files, which have also been added to the repo. The content of the `.report` files contains information about the number of aligned bases, genome identity, etc., for pairwise comparisons, which will be used as a ground truth for testing the `dnadiff` method in `pyani_plus`. 

```bash
/Users/angelikakiepas/Desktop/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0264574.fas /Users/angelikakiepas/Desktop/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0266457.fna
NUCMER

                               [REF]                [QRY]
[Sequences]
TotalSeqs                          1                    1
AlignedSeqs               1 (100.00%)           1 (100.00%)
UnalignedSeqs               0 (0.00%)             0 (0.00%)

[Bases]
TotalBases                     39253                39594
AlignedBases           39169 (99.79%)        39176 (98.94%)
UnalignedBases             84 (0.21%)           418 (1.06%)
```
- I have added `get_dnadiff` to `tools.py`. Although `pyani-plus` will not need `dnadiff`, it was necessary for me to generate fixtures. We can always remove this if it’s not needed. 
- I have also added the directory with all `dnadiff` `.report` files to `conftest.py` fixtures.

### DIFF
--- a/pyani_plus/tools.py
+++ b/pyani_plus/tools.py
@@ -289,3 +289,31 @@ def get_show_diff(cmd: str | Path = "show-diff") -> ExternalToolData:
         raise RuntimeError(msg)
 
     return ExternalToolData(exe_path, version)
+
+
+def get_dnadiff(cmd: str | Path = "dnadiff") -> ExternalToolData:
+    """Return dnadiff and version as a named tuple.
+
+    We expect the tool to behave as follows:
+
+    .. code-block:: bash
+
+        $ dnadiff -V
+            dnadiff
+            DNAdiff version 1.3
+
+    Here the function would return the binary path and "1.3" as the version.
+
+    Raises a RuntimeError if the command cannot be found, run, or if the
+    version cannot be inferred - this likely indicates a dramatically
+    different version of the tool with different behaviour.
+    """
+    exe_path, output = _get_path_and_version_output(cmd, ["-V"])
+
+    match = re.search(r"(?<=version\s)[0-9\.]*", output)
+    version = match.group().strip() if match else None
+    if not version:
+        msg = f"Executable exists at {exe_path} but could not retrieve version"
+        raise RuntimeError(msg)
+
+    return ExternalToolData(exe_path, version)
--- a/tests/generate_fixtures/generate_target_dnadiff_files.py
+++ b/tests/generate_fixtures/generate_target_dnadiff_files.py
@@ -42,6 +42,7 @@ sequences in the query too.
 # Imports
 import shutil
 import subprocess
+import tempfile
 from itertools import permutations
 from pathlib import Path
 
@@ -59,14 +60,7 @@ DELTA_DIR = Path("../fixtures/dnadiff/targets/delta")
 FILTER_DIR = Path("../fixtures/dnadiff/targets/filter")
 SHOW_DIFF_DIR = Path("../fixtures/dnadiff/targets/show_diff")
 SHOW_COORDS_DIR = Path("../fixtures/dnadiff/targets/show_coords")
-
-# dnadiff generates many files; since we only need `.report` files,
-# we will save all files to a temporary location, move the report files
-# to a specific directory, and remove the rest.
-Path("temp_dnadiff").mkdir(parents=True, exist_ok=True)
-
 DNADIFF_DIR = Path("../fixtures/dnadiff/targets/dnadiff_reports")
-DNADIFF_TEMP_DIR = Path("temp_dnadiff")
 
 # Running comparisons
 comparisons = permutations([_.stem for _ in Path(INPUT_DIR).glob("*.f*")], 2)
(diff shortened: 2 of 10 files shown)


pyani_plus/tools.py
```diff
@@ -289,3 +289,31 @@ def get_show_diff(cmd: str | Path = "show-diff") -> ExternalToolData:
...
+    """
+    exe_path, output = _get_path_and_version_output(cmd, ["-V"])
+
+    match = re.search(r"(?<=version\s)[0-9\.]*", output)
+    version = match.group().strip() if match else None
+    if not version:
+        msg = f"Executable exists at {exe_path} but could not retrieve version"
+        raise RuntimeError(msg)
+
+    return ExternalToolData(exe_path, version)
```
peterjc, line 319: Can you add something to ``test_tools.py``, both a good and a bad case using the existing mock commands?

tests/generate_fixtures/generate_target_dnadiff_files.py
```diff
@@ -58,6 +60,14 @@
...
 SHOW_COORDS_DIR = Path("../fixtures/dnadiff/targets/show_coords")
 
+# dnadiff generates many files; since we only need `.report` files,
+# we will save all files to a temporary location, move the report files
+# to a specific directory, and remove the rest.
+Path("temp_dnadiff").mkdir(parents=True, exist_ok=True)
+
+DNADIFF_DIR = Path("../fixtures/dnadiff/targets/dnadiff_reports")
+DNADIFF_TEMP_DIR = Path("temp_dnadiff")
+
```
peterjc, line 70: Maybe use ``import tempfile`` and something like ``generate_anib_blast_files.py`` does for a temp directory?

pyani_plus/tools.py
kiepczi: Sure! I just realised I haven't tested this part of the code!

tests/generate_fixtures/generate_target_dnadiff_files.py
```diff
@@ -71,11 +81,14 @@
     file.unlink()
 for file in SHOW_COORDS_DIR.glob("*.mcoords"):
     file.unlink()
+for file in DNADIFF_DIR.glob("*.report"):
+    file.unlink()
```
peterjc, line 85: Using a temp directory via ``tempfile`` would take care of all this cleanup automatically.

peterjc: Is this "cleanup" before running the tool mean to remove any pre-existing fixtures? This isn't needed if we're going to replace files, but makes sense if we were changing the example set (e.g. alternative genomes).

kiepczi: I was about to ask a question on this. This line removes any pre-existing fixtures, so each time the script is re-run, it deletes the existing fixtures and generates new ones, regardless of whether the example set of genomes has changed or not. This part of code was introduced by someone (I assume either you or @widdowquinn) for other fixtures like `.mcoords` and `.rdiff` files. I simply expanded this logic for `.report`. I'm happy to make changes if needed, but I am a little unsure of what the best practise would be.

Removing this part of the code and re-running it on the same set of genomes doesn't break anything, as it simply overwrites the existing files.

Perhaps keeping this would be a good idea because re-running the scripts, even when the set of input genomes doesn't change, does no harm. However, it might complicate things when we decide to change the set of genomes. What do you think?

tests/generate_fixtures/generate_target_dnadiff_files.py
kiepczi: I have changed it now. Thanks for introducing me to `tempfile`!

tests/generate_fixtures/generate_target_dnadiff_files.py
kiepczi: Here is the commit that introduced the clean up of fixtures when regenerating them 57cdd6fd472c4c9afcae1656893acb09f8a2c32f. Are you still happy to keep it that way :)

peterjc: It was me then. 😩

Yes, I think it makes sense particularly for the use case of regenerating the fixtures as part of changing the input genomes. But we should perhaps add a comment about this (and ought to do this to the other scripts too)?

kiepczi: It makes sense! I have added a comment now.

Regarding the changes to the other script, would you like me to make those changes in this branch or perhaps in a new one?

peterjc: A small separate PR would be safest - and any conflicts wouldn't hold up the dnadiff work

peterjc (approved):

kiepczi:
I will merge this now, so I can work on the `dnadiff` method calculations.

Issue #104, opened 2024-10-04
Compare fastANI output to reference matrices from legacy pyANI


peterjc:
We already have in place fixtures and tests to run fastANI, but don't yet confirm the values at the end match what pyANI v0.3 was producing.

Cross reference #100 which adds the expected values for ANIm as TSV files.

peterjc:
From the fastani output in our current fixtures:

```
$ cat tests/fixtures/fastani/targets/*.fastani | sed "s#\.\./fixtures/viral_example/##g" | xsv table -d "\t"
MGV-GENOME-0264574.fas  MGV-GENOME-0264574.fas  100      13  13
MGV-GENOME-0264574.fas  MGV-GENOME-0266457.fna  99.5014  12  13
MGV-GENOME-0264574.fas  OP073605.fasta          99.9386  13  13
MGV-GENOME-0266457.fna  MGV-GENOME-0264574.fas  99.5247  13  13
MGV-GENOME-0266457.fna  MGV-GENOME-0266457.fna  100      13  13
MGV-GENOME-0266457.fna  OP073605.fasta          99.5129  13  13
OP073605.fasta          MGV-GENOME-0264574.fas  99.8333  13  19
OP073605.fasta          MGV-GENOME-0266457.fna  99.4912  13  19
OP073605.fasta          OP073605.fasta          100      19  19
```

The ANI mapping to the identity table looks sound (other than the floating point annoyance with 99.4912 to 0.9951289999999999 etc):

```
$ xsv table tests/fixtures/fastani/matrices/matrix_identity.tsv 
                                    689d3fd6881db36b5e08329cf23cecdd:1  78975d5144a1cd12e98898d573cf6536:2  5584c7029328dc48d33f95f0a78f7e57:3
689d3fd6881db36b5e08329cf23cecdd:1  1.0                                 0.995014                            0.999386
78975d5144a1cd12e98898d573cf6536:2  0.995247                            1.0                                 0.9951289999999999
5584c7029328dc48d33f95f0a78f7e57:3  0.9983329999999999                  0.9949119999999999                  1.0
```

The coverage estimate from dividing column 4 (orthologous fragments matched) by column 5 (fragments) should be 13/13 or 19/19 or just 1.0 on the diagonal for self-self, otherwise either 12/13 = 0.9230769230769231, or 13/13 = 1.0, or 13/19 = 0.6842105263157895 - which is not what we got from pyANI 0.3

widdowquinn:
[...]

PR #116, opened 2024-10-10
Assorted fixes for testing matrix comparisons

peterjc:
This is work from the ANIb and fastANI matrix testing (issues #102 and #104).

The helper function will now test all available matrices (which we want to do generally), but this currently breaks with fastANI - so that is disabled in the short term.


peterjc:
Codacy is just grumbling about asserts in test code 😞 

I'd like to merge this afternoon, and then I think I can have the fastANI and ANIb matrix testing PRs ready.

peterjc:
Thank you

PR #118, opened 2024-10-10
refractor dnadiff workflow, generate new fixtures, update tests

kiepczi:
Previously, the workflow generated `.rdiff` files for reference calculations. After agreeing that we want to report values for the query instead, I made the following changes:

- Updated the `dnadiff` workflow to generate `qdiff` files instead of `rdiff` files.
- Modified the `generate_fixtures` function to create `qdiff` files.
- Added support for `.qdiff` files in the workflow.
- Changed the arguments for `show-coords`. This was done for easier paring of the files, and so that all arguments are the same as the ones implemented in `dnadiff`.

I would like to merge this PR today (if possible), so I can continue working of the `dnadiff` method :)


tests/fixtures/dnadiff/targets/delta/OP073605_vs_MGV-GENOME-0266457.delta
```diff
@@ -1,4 +1,4 @@
-/Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/OP073605.fasta /Users/pc40583/repositories/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0266457.fna
+/Users/angelikakiepas/Desktop/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/OP073605.fasta /Users/angelikakiepas/Desktop/pyani-plus/tests/generate_fixtures/../fixtures/viral_example/MGV-GENOME-0266457.fna
```
peterjc, line 1: I wouldn't have bothered checking in these meaningless changes to the delta files - it just adds noise to the commit history.

peterjc (approved):
If you are happy with revising a commit (to avoid the near meaningless fixture file changes), then do that first.

Otherwise, I'm not familiar with the parameter changes, but OK to merge.

tests/fixtures/dnadiff/targets/delta/OP073605_vs_MGV-GENOME-0266457.delta
kiepczi: Fair point. Is there a way to keep other changes, but at the same time disregard changes to the delta files?

kiepczi: This has been resolved in https://github.com/pyani-plus/pyani-plus/pull/118/commits/a0fde5448eda119f7dc7dacf5c3ba3fdb0eb0ba5.

kiepczi:
The dnadiff method is a new addition to pyani_plus that was not previously implemented in the legacy pyani. Once I start working on the actual calculations, we will be sure if the workflow generates appropriate outputs :)

I will merge these changes now. Hopefully, they are correct this time.

peterjc:
A teachable moment - no harm done.

PR #123, opened 2024-10-11
Make DB logging API stricter; rename add_comparison -> db_comparison etc

peterjc:
I want to change the command line and internal API to make adding dependent table rows more configurable. In particular currently adding a comparison will implicitly add the configuration and genome rows if not already there - which may be undesirable (e.g. mismatch between the intended configuration already in the DB, and what snakemake is trying to use instead which if wrong will not already exist). This was deliberate while bootstrapping the framework as I could use the log-comparison functionality in isolation, creating a new DB as needed.

With these changes ``log_comparison`` and the ``log_<method>`` functions like ``log_fastani`` will now REQUIRE that the database exists and contains the configuration and genomes already.

This ensures our snakemake workflows WILL FAIL if they are trying to write to a non-existent configuration due to an error in propagating the configuration settings.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #128, opened 2024-10-15
private CLI to record ANIm comparison in the DB

peterjc:
Continues work for #4 to run nucmer for the ANIm method, this expands the private CLI to allow logging a single ANIm comparison -- following the pattern used for ANIb and fastANI.

The final step would be to run this from snakemake (which I have nearly working but am missing something causing cryptic failures).

This also improves the direct test coverage of the functions for logging ANIb and fastANI comparisons which are already used from snakemake.

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review


pyani_plus/private_cli.py
```diff
@@ -426,6 +426,80 @@ def fragment_fasta(
...
+        session,
+        configuration_id=config.configuration_id,
+        query_hash=query_md5,
+        subject_hash=subject_md5,
+        identity=identity,
+        aln_length=r_aligned_bases,
+        sim_errors=sim_errors,
+        cov_query=cov_query,
+        cov_subject=cov_subject,
+    )
```
peterjc, line 497: In order to log the comparison, we (now) must first log the two genomes and the configuration.

Prior to merging #123, we could have just asked to log the comparison, and it would silently have filled in the other tables too.

pyani_plus/private_cli.py
```diff
@@ -426,6 +426,80 @@ def fragment_fasta(
...
+    # Don't use any of fragsize, maxmatch, kmersize, minmatch (configuration table entries)
+) -> int:
+    """Log a single pyANI-plus ANIm pairwise comparison (with nucmer) to the database."""
+    # Assuming this will match as expect this script to be called right
+    # after the computation has finished (on the same machine)
+    nucmer_tool = tools.get_nucmer()
+
+    r_aligned_bases, q_aligned_bases, identity, sim_errors = method_anim.parse_delta(
+        deltafilter
+    )
```
peterjc, line 445: As per legacy pyANI v0.3, should we use the query or reference aligned bases as the alignment length for query-vs-subject?

kiepczi: We had a conversation about this in https://github.com/widdowquinn/pyani/issues/151! Since, the behaviour of nucmer is not symmetrical (eg. values might _slightly_ differ for A:B and B:A companions), we were left with different choices:
1. Calculation of a summary (max, min, mean) between A:B and B:A.
2. Reporting for the query, such as reporting for genome B in A:B comparisons and genome A in B:A comparisons.
3. Reporting for the reference, for example, reporting for genome A in A:B comparisons and genome B in B:A comparisons.

We agreed that there is no right answer, and we have decided to report values for the query in this case.

peterjc: The old code did this:

```python
        (
            query_tot_length,
            subject_tot_length,
            weighted_identity,
            tot_sim_error,
        ) = parse_delta(deltafile)
```

And logged ``query_tot_length`` as the query-vs-subject alignment length, _and_ ``subject_tot_length`` as the subject-vs-query alignment length.

Does the new ``parse_delta`` return query or subject length first?

peterjc: Quoting old pyANI v0.3,

```
def parse_delta(filename: Path) -> Tuple[int, int, float, int]:
    """Return (reference alignment length, query alignment length, average identity, similarity erors)
    ...
    return (raln_length, qaln_length, avrg_ID, sim_error)
```

Yet when used,

```python
        (
            query_tot_length,
            subject_tot_length,
            weighted_identity,
            tot_sim_error,
        ) = parse_delta(deltafile)
```

It does look like query and reference (aka subject) got flipped there?

peterjc: Quoting pyANI-plus,

```
def parse_delta(filename: Path) -> tuple[int, int, float, int]:
    """Return (reference alignment length, query alignment length, average identity, similarity errors).
    ...
    return (
        get_aligned_bases_count(regions_ref),
        get_aligned_bases_count(regions_qry),
        avrg_identity,
        sim_error,
    )
```

So this is reference then query

kiepczi: Well... I'm getting a little bit confused here. I refer to them as reference_vs_query, because that's what nucmer refers to them as. In new `pyANI`, the code reported the values for the query. 

So far, the tests that were carried out all match between old and new pyani, so perhaps there were swapped  somewhere?

peterjc: I _believe_ the tests on this PR confirm we're matching legacy pyANI output, but all the same, this line in old pyANI appears to swap the variable names:

```python
        (
            query_tot_length,
            subject_tot_length,
            weighted_identity,
            tot_sim_error,
        ) = parse_delta(deltafile)
```

We could also have different interpretations of subject vs query and the matrix output (transposed or not)? On the brightside, the expected ANIm alignment length matrix is  asymmetric for the three viruses.

peterjc: 😱

Well, I guess this matches the tested behaviour in pyANI v0.3 and that's OK?

kiepczi: At this point, I’m not sure... Since I started working on pyANI in December, I haven’t touched this part of the code. The only thing I fixed was the issue in parse_delta, where duplicated regions were being counted twice.

When @widdowquinn and I agreed to report values for the query, I assumed he meant the term "query" as used in the nucmer documentation. Honestly, I’m not sure why the old pyANI refers to them as query vs subject. Personally, I’d lean toward using reference vs query—it’s clearer and aligns better with `nucmer` documentation. Sticking to consistent terminology would make things easier for end-users and anyone else working on this project in the future. I can be wrong tho 😭

kiepczi: > Well, I guess this matches the tested behaviour in pyANI v0.3 and that's OK?

The calculations in legacy `pyANI` and `pyANI-plus`, should be correct as I remember calculating the values by hand from `.filter` files, when trying to fix the genome coverage issue. The question is what value we would want to report... `query`, `subject`, or `reference`?

kiepczi: I think the best way to go about this is to report `query` (eg. `query` as in `nucmer` documentation), as this is the sequence that is being compared.

peterjc: NCBI BLAST uses the terms query vs subject, and so currently do our column names in the table. I'm used to having subject interchangeable with reference for pairwise alignments.

kiepczi: My apologies for my clumsy wording here. I absolutely agree with you that using subject interchangeably with reference for pairwise alignment is okay. I just don't think we should refer to it as query vs. subject/reference (when constructing a command in `Snakemake`) in the context of aniM comparisons, since the comparisons are reference/subject vs. query. This could cause issues, like the one we encountered with the old `pyANI`, if that makes sense.

kiepczi: I looked at the `parse_delta` function in `pyANI-plus`, and it reportes the correct values for what is consiedered a `query` and `reference` in `nucmer`. 

If you happy with work done on this branch, feel free to merge this into main.

peterjc: Thank you

PR #137, opened 2024-10-18
Catch empty nucmer file explicitly

peterjc:
Without this we got a divide-by-zero when calculating the average percentage identity. We need a test case with a real nucmer delta file giving no alignment...

[Spotted while working on something else which appears to have accidentally triggered empty delta files. In future work we will want a test case which really does have no alignment, and at that point handle the divide-by-zero situation explicitly]

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review

### DIFF
--- a/pyani_plus/methods/method_anim.py
+++ b/pyani_plus/methods/method_anim.py
@@ -132,7 +132,12 @@ def parse_delta(filename: Path) -> tuple[int, int, float, int]:
     aligned_bases = 0  # Hold a count of aligned bases for each sequence
     weighted_identical_bases = 0  # Hold a count of weighted identical bases
 
-    for line in [_.strip().split() for _ in filename.open("r").readlines()]:
+    # Ideally we wouldn't read the whole file into memory at once...
+    lines = [_.strip().split() for _ in filename.open("r").readlines()]
+    if not lines:
+        msg = f"Empty delta file from nucmer, {filename}"
+        raise ValueError(msg)
+    for line in lines:
         if line[0] == "NUCMER":  # Skip headers
             continue
         # Lines starting with ">" indicate which sequences are aligned
--- a/tests/test_anim.py
+++ b/tests/test_anim.py
@@ -153,6 +153,9 @@ def test_delta_parsing(anim_nucmer_targets_filter_indir: Path) -> None:
         222,
     )
 
+    with pytest.raises(ValueError, match="Empty delta file from nucmer, /dev/null"):
+        method_anim.parse_delta(Path("/dev/null"))
+
 
 def test_aligned_bases_count(aligned_regions: dict) -> None:
     """Check only aligned bases in non-overlapping regions are counted."""


pyani_plus/methods/method_anim.py
```diff
@@ -132,7 +132,12 @@ def parse_delta(filename: Path) -> tuple[int, int, float, int]:
     aligned_bases = 0  # Hold a count of aligned bases for each sequence
     weighted_identical_bases = 0  # Hold a count of weighted identical bases
 
-    for line in [_.strip().split() for _ in filename.open("r").readlines()]:
+    # Ideally we wouldn't read the whole file into memory at once...
+    lines = [_.strip().split() for _ in filename.open("r").readlines()]
+    if not lines:
+        msg = f"Empty delta file from nucmer, {filename}"
+        raise ValueError(msg)
+    for line in lines:
```
kiepczi, line 140: I'm afraid this will not work, as `nucmer` does not simply return an empty file if there is no alignment. It still provides two lines: the first line shows the genomes used [REF] and [QRY], and the second line indicates which program was run, e.g., NUCMER or PROMER. See below:

```bash
/Users/angelikakiepas/Desktop/viral_example/no_alignment/MGV-GENOME-0264574.fas /Users/angelikakiepas/Desktop/viral_example/no_alignment/MGV-GENOME-0357962.fna
NUCMER
```

Maybe an alternative solution would be something like this for line 169:

```python
try:
    avrg_identity = weighted_identical_bases / aligned_bases
except ZeroDivisionError:
    avrg_identity = NULL # or 0 or NA or NaN
```

kiepczi: If you need genomes that return no aligment, here it is:
[no_aligment.zip](https://github.com/user-attachments/files/17436732/no_aligment.zip). Or if you want I can take care of this, because this might also be an issue for `dnadiff` method too.

peterjc: I have a commit very much like that here too 😁

But until it has a test case, I was holding off - in part because zero vs NaN vs NA is up for discussion here.

I wanted to think about where to put problem alignments like this in the tests (and have some ideas for other failure modes).

kiepczi: Now thinking about this... Your code might address a different issue. For example, if `nucmer` fails for any reason and does not write anything to a file at all, your code should catch this 😄 So, if you want to leave this that's totally fine, but this is unlikley to fix `ZeroDivisionError`.

peterjc: And back to this fix - the point was even on a bad alignment I was expecting ``nucmer`` to output something, even if just the header line.

I want to give a clear exception if it is literally an empty file - which should be treated as an error condition. I had messed up the workflow file command and was accidentally generating empty files while working on #138.

(Note for some formats an empty file cannot be distinguished from no alignments, BLAST TSV output in particular comes to mind)

kiepczi: > But until it has a test case, I was holding off - in part because zero vs NaN vs NA is up for discussion here.


I vaguely remember from our meeting at some point that we agreed on using `NULL`, though I can't recall exactly why. I'm happy to have this discussion again. I'm not a mathematician by any means, so I'd better stay 🤫.

kiepczi: > And back to this fix - the point was even on a bad alignment I was expecting `nucmer` to output something, even if just the header line.
> 
> I want to give a clear exception if it is literally an empty file - which should be treated as an error condition. I had messed up the workflow file command and was accidentally generating empty files while working on #138.
> 
> (Note for some formats an empty file cannot be distinguished from no alignments, BLAST TSV output in particular comes to mind)


Ah! I got you now! I thought this was to propose a fix for `ZeroDivisionError`. I must have misunderstood your commit message. In this case, I'm happy for this to be merged into the main branch.

peterjc: We should discuss and document this.

PR #138, opened 2024-10-18
Silence stdout during snakemake workflows

peterjc:
This PR runs snakemake in quiet mode (default is quite verbose), and since snakemake doesn't do it for us, also silences the underlying command line tools as well (which may not be necessary when run via a cluster).

The tool silencing is being done using the venerable Perl tool ``chronic`` (originally invented for use with ``cron``), which will capture and suppress the stdout/stderr provided the tool has a zero return code. If the tool fails, the output is restored.

This does add another dependency and point of failure, so I would prefer snakemake handled it - see https://github.com/snakemake/snakemake/issues/3152

Note that ``chronic`` is part of the ``moreutils`` package which is in conda-forge, but at the time of writing has an undeclared perl dependency provided by the ``perl-ipc-run`` package which is currently in bioconda only. That in turn needs ``perl-io-tty`` which is in conda-forge, but not yet packaged for ARM. See https://github.com/conda-forge/moreutils-feedstock/issues/6

A possible alternative is a new chronic implementation in go, which is not yet packaged for conda - https://github.com/docwhat/chronic

The motivation is for the main public facing command line interface (CLI), where I intend to use a progress bar alone (which with these changes works nicely in the non-error case).

## Type of change

## Action Checklist

- [x] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Work on this branch
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Commit branch
- [x] Request a code review

### DIFF
--- a/pyani_plus/snakemake/snakemake_scheduler.py
+++ b/pyani_plus/snakemake/snakemake_scheduler.py
@@ -25,6 +25,7 @@ import importlib.resources as impresources
 from pathlib import Path
 
 from snakemake.api import ConfigSettings, DAGSettings, ResourceSettings, SnakemakeApi
+from snakemake.settings.types import OutputSettings, Quietness
 
 from pyani_plus import workflows
 
@@ -75,7 +76,7 @@ class SnakemakeRunner:
         snakefile = impresources.files(workflows) / self.workflow_filename
 
         # Use the defined workflow from the Python API
-        with SnakemakeApi() as snakemake_api:
+        with SnakemakeApi(OutputSettings(quiet={Quietness.ALL})) as snakemake_api:
             workflow_api = snakemake_api.workflow(
                 snakefile=snakefile,
                 resource_settings=ResourceSettings(cores=config_args["cores"]),
--- a/pyani_plus/workflows/snakemake_anib.smk
+++ b/pyani_plus/workflows/snakemake_anib.smk
@@ -45,7 +45,7 @@ rule fragment:
     output:
         "{outdir}/{genomeA}-fragments.fna",
     shell:
-        ".pyani-plus-private-cli fragment-fasta --outdir {wildcards.outdir} --fragsize {params.fragLen} {input.genomeA}"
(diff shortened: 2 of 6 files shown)


peterjc:
GitHub Actions failures under Linux:

```
BEGIN failed--compilation aborted at /home/runner/miniconda3/envs/pyani-github/bin/chronic line 37.
```

Says is it using:

```
moreutils-0.5.7            |                1          27 KB  bioconda
```

I did test this on our Linux server and its worked, but here I have version 0.67 from conda-forge. In my local copies of the script, line 37 is still part of the help text.

widdowquinn:
The response at https://github.com/snakemake/snakemake/issues/3152 seems to have misunderstood your issue, and to think you want to log errors using `echo` <shrug />.

widdowquinn:
Is it possible to avoid the additional dependency by wrapping calls to each tool in a function using `subprocess.run()`? I'm thinking that the function might:

1. checks the return code
2. log stdout/stderr to a `pyani-plus` logger if the return code is non-zero

That would give us flexibility to define return codes that are nonero but not in error, for any non-standard tools we might come across.

widdowquinn:
Under https://docs.python.org/3/library/subprocess.html there is the note:

> If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.

Can the main script see errors/logs from the workflow scripts?

widdowquinn (commented):
Maybe we can avoid the additional dependency? (see notes above)

widdowquinn:
Let's merge this in the interim, but with a view to testing the feasibility of the `subprocess.run(check=True)` -style wrapper, which may let us log tool run failures gracefully in the `pyani-plus` logger.

PR #148, opened 2024-10-23
similarity errors (dnadiff method) 

kiepczi:
__Proposed changes__
This PR:
- adds calculations for similarity errors in the `dnadiff` method
- generates `matrix_sim_error.tsv` with expected values
- tests the values against `matrix_sim_error.tsv`

__Backgroud__
At the last meeting, we discussed how we would like to calculate similarity errors for the `dnadiff` method. For documentation purposes, let me explain a small example here.

Let's say we have the following alignment:
```
A T T C G T T - G 
| | |     | |   | 
T C T - - C T C G 
```

for which the similarity error is 3:
```
A T T C G T T - G 
| | |     | |   | 
T C T - - C T C G
    *       *   *   <- 3 identical bases (identity = 0.5 %)
      * *     *     <- 3 gaps
* *       *         <- 3 similarity errors
```
In the `dnadiff` method, we do not have information about the number of identical bases, but we do have the identity percentage. Therefore, we can reverse calculate the similarity errors as follows:

```
aligned_bases_no_gaps - (aligned_bases_no_gaps  * identity)

6 - (6 * 0.5) = 3
```

This got a little bit more complicated when comparing the "expected" values and those reported by `pyANI-plus`, due to differences in rounding errors.

In this example, let's consider the pairwise comparison between MGV-GENOME-0266457 (query) and MGV-GENOME-0264574 (subject).

When generating the expected values from `.report` files, we are forced to use % identity rounded to the 2nd decimal place (on a scale of 0-100), which gives us a slightly lower similarity error than the one we would report, as we do not round % identity in our calculations in `pyANI-plus`.

`.rdiff` file:
```
                               [REF]                [QRY]
[Sequences]
TotalSeqs                          1                    1
AlignedSeqs               1(100.00%)           1(100.00%)
UnalignedSeqs               0(0.00%)             0(0.00%)

[Bases]
TotalBases                     39253                39594
AlignedBases           39169(99.79%)        39176(98.94%)
UnalignedBases             84(0.21%)           418(1.06%)

[Alignments]
1-to-1                             2                    2
TotalLength                    59174                59187
AvgLength                   29587.00             29593.50
AvgIdentity                    99.63                99.63

M-to-M                             2                    2
TotalLength                    59174                59187
AvgLength                   29587.00             29593.50
AvgIdentity                    99.63                99.63
```

Calculations from `.report`:
```
39176 - (39176 * (99.63/100)) = 144.95120000000315
```

Calculations in `pyANI-plus`:
```
39176 - (39176 * 0.996266174669021) = 146.27634116643458
```
This gives us a diffrence of approx. 1.33. 

For this reason, when comapring the dnadiff sim error matrices, I set up the `absolute_tolerance` to 1.33 (see below). But I'm not sure if that's a good practice.

```python
    if (matrices_path / "matrix_sim_errors.tsv").is_file():
        compare_matrix(run.sim_errors, matrices_path / "matrix_sim_errors.tsv")
        if "dnadiff" in str(matrices_path):
            compare_matrix(
                run.sim_errors,
                matrices_path / "matrix_sim_errors.tsv",
                absolute_tolerance=1.33,
            )
            checked = True
        else:
            compare_matrix(run.sim_errors, matrices_path / "matrix_sim_errors.tsv")
        checked = True
```

The only alternative I can think of is to round the identity when calculating the similarity errors in `pyANI-plus`, which I am not a big fan of, as this might make the calculations less accurate (if I understand the concept of similarity errors correctly).

I was also wondering if I should change the values in` matrix_sim_errors.tsv` to integers rather than use floats.


pyani_plus/private_cli.py
```diff
@@ -625,7 +625,11 @@ def log_dnadiff(
         subject_hash=subject_md5,
         identity=identity,
         aln_length=aligned_bases_with_gaps - gap_lengths,
-        sim_errors=None,  # Leaving this as None for now (How should we calculate this?)
+        sim_errors=int(
```
peterjc, line 628: Do we want to round down (which is what ``int`` will do), or round to nearest whole number (with ``round``)?

peterjc:
Short of generating the ``.report`` file and parsing that, this seems as good as we can get.

Why don't we run ``dnadiff`` tool directly? It is significantly slower than the approach currently implemented?

The PR looks fine to me, but I think we'll need Leighton to rule on this (e.g. should we keep sim-errors as integers in the database).

pyani_plus/private_cli.py
kiepczi: I just used `int` for now, but since I'm not a mathematician not sure if `round` would be more appropriate. Perhaps, leaving this until monday won't do any harm.

kiepczi:
> Why don't we run `dnadiff` tool directly? It is significantly slower than the approach currently implemented?
I'm not sure how much quicker it is to run dnadiff instead of just generating the files that we need. We decided not to run dnadiff, as it generates 11 files, and we only really need 4. I understand that we can remove the files that are not needed so it does not generate a lot of unnecessary files and take up space on the disk for larger analyses, but they round their numers for identity and coverage, and perhaps we want to be more accurate?

peterjc:
All the files are being generated in temporary working directories anyway, so cleaning up isn't a big issue. I would suggest some timings on bacterial genomes might be informative.

It may be we can run ``dnadiff`` and use its report for sim-errors, but for identity and coverage extract them from the intermediate files (as done now)?

kiepczi:
As for reporting values in pyANI-plus, I think there is a shorter way to do this, and there is no need to run dnadiff. We can simply report sim_error values by using the calculated identity rounded to four decimal places from the intermediate files.

For example instead of calcuating sim errors in this manner:
```
39176 - (39176 * 0.996266174669021) = 146.27634116643458
```
we simply do:

```
39176 - (39176 * round(0.996266174669021,4)) = 144.95120000000315
```

kiepczi:
Running pairwise comparisions on two viral genomes (MGV-GENOME-0264574 vs MGV-GENOME-0266457) with dnadiff takes:
```
bash dnadiff_viral.sh  0.24s user 0.11s system 93% cpu 0.371 total
```
where calling only programs that we need (`nucmer`, `delta-filter`, `show-diff` and show-coords) takes:

```
bash intermediate_viral.sh  0.18s user 0.07s system 95% cpu 0.266 total
```

The diffrence is much larger for much bigger bacterial genomes ( GCF_014207295.1 vs GCF_014202475.1). Time taken for dnadiff:

```
bash dnadiff_bacerial.sh  22.00s user 0.41s system 99% cpu 22.425 total
```

Time taken to generate only intermediate files:

```
bash intermediate_bacterial.sh  19.57s user 0.29s system 100% cpu 19.844 total
```

peterjc:
So its about a 10% overhead to call dnadiff? Personally I don't find the speedup alone compelling enough to justify the complexity in having to replicate their tool's internal logic. But I am likely missing other factors in this discussion.

pyani_plus/private_cli.py
widdowquinn: We should use `round()` to get the closest integer.

widdowquinn:
> So its about a 10% overhead to call dnadiff? Personally I don't find the speedup alone compelling enough to justify the complexity in having to replicate their tool's internal logic. But I am likely missing other factors in this discussion.

If we're already using the intermediate files for identity/coverage, then we could use them for sim_errors, too. `dnadiff` doesn't report sim_errors directly, and I don't think they intend their percentages to be used that way, so I'd prefer our approach from the intermediate files.

widdowquinn:
> For example instead of calcuating sim errors in this manner:
> 
> ```
> 39176 - (39176 * 0.996266174669021) = 146.27634116643458
> ```
> 
> we simply do:
> 
> ```
> 39176 - (39176 * round(0.996266174669021,4)) = 144.95120000000315
> ```

Is there a reason you're not using the equivalent:

```
aligned_bases * (1 - id_percentage)
```

widdowquinn (approved):
Great - thanks, @kiepczi !

PR #160, opened 2024-10-29
Drop unused maxmatch from DB and CLI

peterjc:
In discussion with @kiepczi last week, we we unsure that we needed to keep the ``maxmatch`` parameter in the database.

This was used in legacy pyani to control how MUMmer was used (and indeed our dnadiff method does run nucmer with --maxmatch), but is not currently exposed as a user setting.

Currently this field is an optional boolean in the DB, i.e. in Python speak it can be True, False, or None. This is odd, especially for setting this at the command line.

It might make sense to replace with a more general ``mode`` string parameter, but this PR simply removes ``maxmatch``.

## Type of change

## Action Checklist

- [x] Work on this branch
  - [ ] new code is commented, especially in hard-to-understand areas
- [x] Commit branch
- [x] Request a code review


widdowquinn:
I'm not convinced about the rationale for dropping `maxmatch` from both the database and CLI.

## A brief history of ANIm options

Historically, the original `pyani` script runs ANIm using MUMmer's `-mum` argument. This computes only matches that are unique in ech of the reference and query sequences.

The `--maxmatch` option was offered at the command line to enable users to override this option, enabling MUMmer's `-maxmatch` argument, which calculates _all_ maximal matches, regardless of uniqueness in each sequence. This is expected to result in different output.

This arrangement is also the case in `pyani` v0.3.

The `--mum` option was probably chosen (my memory is cloudy and I don't think I wrote it down) on the basis of the MUMmer documentation, which has:

> Setting either the `-mum` or `-mumreference` option can significantly cut down on the number of repeat induced matches as opposed to `-maxmatch`, and is recommended for most all applications.

I think this is a sound reason to prefer the `--mum` option in ANIm, but the option was made available to the user.

## `maxmatch` as a database field

The `maxmatch` field in the database for v0.3 was meant to record - only for ANIm - whether the `--maxmatch` option was set at the CLI. It has no role in the other analyses offered in that version of the tool.  I see the argument for this being `True` or `False` (with `False` being the case when using ANIb, for instance), but there's an argument for `None` reflecting a null value in the database to explicitly represent cases where it is not an option and avoid confusion about it being set to "False" when it is not a relevant parameter.

However, when recording the full command line either for `pyani` or for the underlying tool, I can see why it would not need to be a field at all.

## ANIm _vs_ dnadiff

I don't see why we should change ANIm behaviour in this version of `pyani`, but I'm happy to hear arguments for why we should. As a default I think we should continue with the same methodology when it's not demonstrably in error. The original description of ANIm is not detailed (see [https://pmc.ncbi.nlm.nih.gov/articles/PMC2776425/](https://pmc.ncbi.nlm.nih.gov/articles/PMC2776425/)) and silent on the choice of `-mum` _vs_ `-maxmatch` - but I suspect that `-mum` would have been used if the documentation was consulted.

From inspection of the MUMmer source, `dnadiff` _does_ use the `--maxmatch` option with `nucmer`, but it also carries out a golden path analysis that isn't part of the ANIm specification. This can presumably be trusted to find a single unique path that best represents the pairwise alignment, so the `--mum` option may not be required to restrict the original alignment to unique matches only, avoiding repeats.

I think it's find to have `--maxmatch` in the `dnadiff` algorithm, but not in the ANIm algorithm.

## tl;dr

1. ANIm should continue to use `--mum` by default and offer the `--maxmatch` option in the CLI.
2. I'm happy to drop the `maxmatch` database field if there's a better way to record the option.

But I'm open to argument.

pyani_plus/private_cli.py
```diff
@@ -147,7 +146,6 @@ def log_configuration(  # noqa: PLR0913
         program=program,
         version=version,
         fragsize=fragsize,
-        maxmatch=maxmatch,
```
widdowquinn, line 150: `-mum` should be the default option for ANIm for consistency with previous versions and as suggested by MUMmer docs. We can either drop `-maxmatch` as an option, with `-mum` as the default behaviour, or we can keep it.

widdowquinn (changes requested):

peterjc:
1. Currently the code does not use the ``--maxmatch`` option, so that is effectively a bug to be fixed. Doing to and making the ANIm configurations record True or False, and other methods record None makes sense. New issue to follow.


2. I would prefer to replace the optional boolean "maxmatch" (which is very nucmer specific) with a string parameter, say "mode" which seems potentially applicable to other methods in the future. For ANIm this would be ``pyani-plus anim --mode mum ...`` (default) or ``pyani-plus anim --mode maxmatch ...``

This PR can be closed.

Issue #154, opened 2024-10-26
Add resume command to public CLI


peterjc:
Any of the methods will take a non-trivial time to run, so there is a chance of interruption. Perhaps more so with a cluster.

The database design avoids recomputing any pairwise ANI previously logged, so starting a new run will effectively resume (assuming same command is executed, via the shell history perhaps).

I am considering a new resume command taking only the database path and optionally run-id, defaulting to the most recent run. This will then check if any comparisons are missing, and if so run those after checking the command line dependency version matches that from the old run.

widdowquinn:
I like this idea - I'm curious about how it might cascade with runs that refer to other incomplete runs… it's probably more straightforward that I imagine, just now.

peterjc:
Currently if you did the following (where all three examples have the same configuration and thus will share comparisons):

1. Ran ``pyani-plus anib --database demo.db --create-db --name "Run 1" my_genomes/`` but aborted it almost immediately.
2. Ran ``pyani-plus anib --database demo.db --name "Run 2" my_genomes/`` and let it partly complete, and then abort.
3. Run ``pyani-plus list-runs --database demo.db`` will show both run 1 and run 2 have been partially completed.
4. Run ``pyani-plus anib --database demo.db --name "Run 3" my_genomes/`` and let it finish
5. Run ``pyani-plus list-runs --database demo.db`` will show all the runs are complete, but only "Run 3" will say this in its status column.

New plan:

1. Ran ``pyani-plus anib --database demo.db --create-db --name "Run 1" my_genomes/`` but aborted it almost immediately.
2. Run ``pyani-plus resume --database demo.db`` and let it partly complete then abort
3. Run ``pyani-plus list-runs --database demo.db`` will show only run 1 exists and has been partially completed.
4. Run ``pyani-plus resume --database demo.db`` and let it finish
5. Run ``pyani-plus list-runs --database demo.db`` will show just "Run 1" and it will completed in its status column.

peterjc:
I have this working with one caveat - it needs to know the input FASTA directory name. It makes sense to me to record this in the run table...

PR #158, opened 2024-10-28
SLURM support

peterjc:
Revised version of #149 towards #132. Does not have any new tests yet...

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #162, opened 2024-10-30
Use the ORM config names consistently

peterjc:
This is work in aid of the resume functionality planned for #154, where it helped to have all the methods use the same variable names for k-mer length, fragment size, etc.

They are now standardized on the names used in our database.

These can then be passed to snakemake centrally rather than making the parameter/config dict separately for each method.


peterjc:
This is giving me conflicts for both #163 and #158 so I'm going to merge it now...

Issue #175, opened 2024-11-10
Simplifiy private CLI for logging ANI results


peterjc:
Currently each private ANI method takes the filename(s) to parse, the query and subject FASTA filenames, the DB to write to, and the configuration values (like k-mer size etc).

Rather than the configuration values (which are fiddly and change between tools), we can just specify the DB configuration ID, or simply the DB run ID. The later would also allow looking up the FASTA filenames to get the MD5 checksums, rather than recomputing them (again), so should be faster.

peterjc:
This API change enables bulk logging of multiple ANI values at once, needed for faster sourmash (done in PR #177) and row-wise fastANI (issue #180).

PR #177, opened 2024-11-11
Build a single sourmash sketch per genome

peterjc:
Previously was building a double genome sketch, and giving this to the compare command. Now we buid a single sketch per genome, and when we do the compare, give it two sketchs to compare. This means we just need *N* smaller sketches, rather than *N^2* double sketches.

Also now listing A then B (but seems fully symmetric anyway).

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review


peterjc:
Indicative raw timings with 200 bacteria, nearly 6 mins to build the 200 sig k=51 files with ``sourmash sketch``, and then only about 12s to build the ANI matrix in single call to ``sourmash compare`` (!).

Even with 465 signatures, only needed just under a minute to output an ANI matrix. For 657, almost 2 mins. For 871, about 3.5mins.

We _really_ don't want to be making command line calls to ``sourmash compare`` for each pair, even if sub-second! The overhead may be negligible with slower methods like ANIm or ANIb, but would best be avoided here.

peterjc:
The approach used here for logging the all-vs-all sourmash output to the DB in a single call is via the DB run-id as proposed on #175 and implemented on #176

peterjc:
Note that since the sourmash CLI is written in Python, there is the possibility to call the internals of ``sourmash compare`` directly in order to have a working progress bar. That might be worth considering as a followup PR, but does have downsides/risks.

peterjc:
NOTE: I would prefer to merge #176 before this, and have another branch dealing with the merge conflicts from that: ``sourmash_and_run_id`` which I can use to update this PR.

pyani_plus/workflows/snakemake_sourmash.smk
```diff
@@ -42,32 +42,31 @@ rule sketch:
...
+        run_id=config["run_id"],
         outdir=config["outdir"],
         mode=config["mode"],
         kmersize=config["kmersize"],
         extra=config["extra"],  # This will consist of either `scaled=X` or `num=X`.
     input:
-        "{outdir}/{genomeA}_vs_{genomeB}.sig",
-        genomeA=get_genomeA,
-        genomeB=get_genomeB,
+        expand("{{outdir}}/{genome}.sig", genome=sorted(indir_files)),
```
widdowquinn, line 60: This gets used in the command line as a list of that many filenames… is there a buffer size upper limit that we might encounter here?

widdowquinn: Maybe write these filenames to a file and use redirection? (does this still potentially hit the limit)?

widdowquinn (approved):
Incredible speed improvements! Just need to see how divergent the results are ;)

peterjc:
Rebased onto main, after applying the logging changes etc.

PR #195, opened 2024-11-27
Reporting None for comparisons with no alignments

kiepczi:
This partially addresses issue #140.

We show now handle comparisons that return no alignments by setting all values (e.g., sim errors, identity, coverage, alignment length, etc.) to None, rather than raising different types of errors.

Unfortunately, t took me longer than expected. I run into a small issue with the `dnadiff` method failing. I hadn't realised that in some cases, even when alignments are found, `.qdiff` files can be empty because there are "no gaps" in the alignments. Additionally, `Ruff` flagged some style issues in the code, so I had to come up with a different solution.

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [X] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [X] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [X] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [X] Create a new branch with a short, descriptive name
- [X] Work on this branch
  - [X] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [X] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [X] tests for the change have been added that demonstrate the fix or feature works
- [X] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [ ] Rebase against `origin/master`
- [X] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [X] Commit branch
- [X] Submit pull request, describing the content and intent of the pull request
- [X] Request a code review
- [X] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #194, opened 2024-11-26
Move intermediate workflow files under parent example folder - Issue 192

peterjc:
Follows on from #193 (matrices) to close #192.

In addition to moving the files, we now sort the fastANI queries by filename for the all_vs_subject intermediate files.

Also, do we need to keep the dnadiff ``*.reports`` files in version control? They are only used to produce the expected output matrix, and do not seem to be fully reproducible? At least, when I regenerate them they seem to change a lot.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [ ] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #197, opened 2024-11-29
test matrices output for bad_alignments example (dnadiff and ANIm)

kiepczi:
Since we now report `None` for comparisons that result in no alignment, in this PR I focused on testing the matrices output for `bad_alignments`. This was done as a follow-up to PR #195.

NOTE: `ANIm` matrices for `viral_example` are generated using the old `pyANI`. This approach cannot be used for `bad_alignments` because the old `pyANI` cannot complete the comparisons and instead throws a `ZeroDivisionError`.

For self-to-self comparisons, we know the following:
- identity and coverage values will be 1.0.
- sim_errors will be 0.
- alignment lengths will equal the length of the sequences.

For comparisons other than self-to-self, we expect all values to be represented as `NULL` in the database.

Since all matrices in this case will be identical to the dnadiff matrices, I decided not to write an additional script to generate data. Instead, I copied the dnadiff matrices and named them as ANIm matrices (see below). If this approach should be avoided, I am open to suggestions.

```
	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_sim_errors.tsv ./ANIm_sim_errors.tsv
	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_identity.tsv ./ANIm_identity.tsv
	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_coverage.tsv ./ANIm_coverage.tsv    
	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_aln_lengths.tsv ./ANIm_aln_lengths.tsv
```

## Type of change

## Action Checklist

- [X] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [X] Work on this branch
- [X] Commit branch
- [X] Request a code review


peterjc:
I wouldn't use copies, but symlinks? And for that change I would rewrite the history for this change (this would be important for big files not to swell the history and thus repository size).

Makefile
```diff
@@ -41,9 +41,14 @@ fixtures:
...
-	cd tests/generate_fixtures/; ./generate_target_dnadiff_matrices.py
+	cd tests/generate_fixtures/; ./generate_target_dnadiff_matrices.py ../fixtures/viral_example ../fixtures/viral_example/matrices
+	cd tests/generate_fixtures/; ./generate_target_dnadiff_matrices.py ../fixtures/bad_alignments ../fixtures/bad_alignments/matrices
 	cd tests/generate_fixtures/; ./generate_target_sourmash_files.py
-	cd tests/generate_fixtures/; ./generate_target_sourmash_matrices.py
+	cd tests/generate_fixtures/; ./generate_target_sourmash_matrices.py 
+	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_sim_errors.tsv ./ANIm_sim_errors.tsv
+	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_identity.tsv ./ANIm_identity.tsv
+	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_coverage.tsv ./ANIm_coverage.tsv    
+	cd tests/fixtures/bad_alignments/matrices/; cp ./dnadiff_aln_lengths.tsv ./ANIm_aln_lengths.tsv
```
peterjc, line 51: e.g.

```
ln -s ./dnadiff_aln_lengths.tsv ./ANIm_aln_lengths.tsv
```

tests/fixtures/bad_alignments/intermediates/dnadiff/MGV-GENOME-0264574_vs_MGV-GENOME-0357962.report
```diff
@@ -39,49 +39,49 @@ TandemInsAvg                    0.00                 0.00
...
 A.                          0(0.00%)             0(0.00%)
+C.                          0(0.00%)             0(0.00%)
+T.                          0(0.00%)             0(0.00%)
+G.                          0(0.00%)             0(0.00%)
+.G                          0(0.00%)             0(0.00%)
 .T                          0(0.00%)             0(0.00%)
-.A                          0(0.00%)             0(0.00%)
 .C                          0(0.00%)             0(0.00%)
-.G                          0(0.00%)             0(0.00%)
+.A                          0(0.00%)             0(0.00%)
```
peterjc, line 87: Why does this file (and one other report) change so much? I was seeing this myself when I moved all the files on #194.

peterjc:
Open question: How should the failed pairwise comparisons appear in the exported matrix TSV file on disk? i.e. Null in the DB, None in Python, currently empty string in TSV. Should we be more explicit and write Null in the TSV file?

I would be guided by expectations in downstream analysis packages that people might use to read the matrices, which could include Python, Python with numpy, Python with pandas, R, and Excel

tests/fixtures/bad_alignments/intermediates/dnadiff/MGV-GENOME-0264574_vs_MGV-GENOME-0357962.report
kiepczi: I believe the files contain the same information overall; only the order in which each SNP is reported differs.

kiepczi: I just generated the `.report` files for now, but as you suggested at some point, it might be worth deleting them after we generate the matrices, since we don't use `.report` files for anything else. But I think, it would be best to do it on a diffrent branch.

peterjc: OK, but then in future I don't think we don't really need/want to check this pointless change into the git history.

FYI: To fix this you'd need to rewrite the branch history (not worth the hassle?), or we could apply this PR as a single squashed-commit.

peterjc: Or delete them - even tidier 😉

kiepczi:
That's a good question 🤔 Personally I would lean towards `NULL`, but I don't have a good argument for this.

Makefile
kiepczi: Quick two questions:
1. Why is `ln -s` a better option than `cp` in this case? I'm happy to change it, just asking out of curiosity.
2. I'm aware of re-writing history and how it can be useful in some cases, but I don't really understand the process. I don't want to make even more of a mess. Would you mind going over this with me on Monday? This PR can wait anyway.

peterjc: In this case the files are small, so the disk saving is marginal. However having this as a sym-link makes it very explicit that the contents of ``ANIm_aln_lengths.tsv`` are identical to  ``dnadiff_aln_lengths.tsv``.

I'd be happy to do a live git workthough on this with you on Monday, using ``git rebase origin/main -i`` with the ``-i`` for interactive.

widdowquinn:
`NA` maybe?

widdowquinn:
I'm easy so long as we're consistent. `pandas` reports this list of defaults at https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html from which we could take our pick:

> “ “, “#N/A”, “#N/A N/A”, “#NA”, “-1.#IND”, “-1.#QNAN”, “-NaN”, “-nan”, “1.#IND”, “1.#QNAN”, “<NA>”, “N/A”, “NA”, “NULL”, “NaN”, “None”, “n/a”, “nan”, “null “

widdowquinn:
The `R` tidyverse seems to use `na = c("", "NA")` (https://readr.tidyverse.org/reference/read_delim.html)

peterjc:
Rebased onto latest main, and copied matrices changed to symlinks.

PR #211, opened 2024-12-10
Record sourmash query-containment as query-coverage

peterjc:
Changes our sourmash and branchwater wrappers to always record max-containment as the ANI (which was the default mode), and record containment via the query-coverage matrix along side it.

To do this via ``sourmash compare`` (our ``sourmash`` method), we no longer run sourmash in ``--max-containment`` mode, but just in ``--containment`` mode and infer the max value for each query/subject pair.  This drops ``--mode max-containment`` or ``--mode containment`` as a user-facing option.

For our ``branchwater`` method, this just looks at an additional column from the output.

This PR also transposes the containment values from ``sourmash.csv`` before loading it as our query-coverage values. See discussion on #202 which this PR replaced, and the test case introduced on #212 which confirms this orientation is consistent.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [x] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #212, opened 2024-12-12
Failed alignments and query coverage with artifical test case

### LINKED ISSUE #140: How to report no viable alignment between two genomes?
We discussed this briefly in person on Monday 14 October, but to put this in writing:

The details differ but most if not all of the ANI methods have some threshold below which they find no meaningful alignments, and thus we might consider that pair of genomes to have zero ANI. However, this could equally be regarded as a null result, perhaps the float NaN (e.g. ``numpy.nan``) from zero matching bases divided by zero aligned bases. See also https://en.wikipedia.org/wiki/NaN

The same logic may apply to the other floating point matrices like coverage and hadamard, while for the integer matrices of alignment length and sim-errors, we'd probably just use zero.

We have some "real" examples of this, but equally artificial test cases are easily created (e.g. a single short sequence). This is an important corner case for testing, as some tools will produce empty files meaning no alignments (e.g. BLAST tabular output, or fastANI tabular output), while others will produce a minimal but not empty file. For the later, an actual empty file is a clear cut error - e.g. nucmer and #137.

This issue can be closed once we have tests covering this for all the methods (this need not be a full matrix, probably a pairwise test parsing a fixture from attempting to align two tiny sequences is enough).

peterjc:
(This is a resurrected version of #210, which GitHub didn't let me reopen)

This closes #140 by fixing recording failed alignments from ANIb (were zero), fastANI (were missing), and branchwater (were missing).

It adds a simple artificial test caseof a small contig, a large contig, and both contigs. The small and large contig have nothing in common, so barring noise in the self-vs-self we can expect identity values of ~100% or null (between the small and large):

```
identity:
        small  both   large
small [[~100%, ~100%,  null],
both   [~100%, ~100%, ~100%],
large  [ null, ~100%, ~100%]]
```

From the genome sizes, small is 30% of both together, and large is 70% of both together. Thus for coverage we expect:

```
query-coverage:
        small  both   large
small [[~100%, ~100%,  null],
both   [ ~30%, ~100%,  ~70%],
large  [ null, ~100%, ~100%]]
```

i.e. When the query is both, and the subject is the small contig, we get a perfect match but for only 30% of the query.

Conversely, when the query is the small contig, and the subject if both contigs, we again get a perfect match but it is 100% of the query.

I will use this test case for the sourmash/branchwater plan to record query/subject containment as query coverage, and max-containment as identity - see #211.

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review


pyani_plus/methods/method_anib.py
```diff
@@ -79,7 +79,7 @@ def fragment_fasta_file(
         raise ValueError(msg)
 
 
-def parse_blastn_file(blastn: Path) -> tuple[float, int, int]:
+def parse_blastn_file(blastn: Path) -> tuple[float | None, int | None, int | None]:
```
kiepczi, line 82: Would it be more easier to read if we change it to:

```python
def parse_blastn_file(blastn: Path) -> tuple[float, int, int] | tuple[None, None, None]:
```

kiepczi (commented):
Looks good @peterjc! There is only one small suggestion that I had, but if you feel it is not necessary and perhaps make things harder to read, feel free to ignore it :)

pyani_plus/methods/method_anib.py
peterjc: Easier to read, yes - and I tried that initially.

However, ``mpypy`` complains. I think it _should_ be possible with some refactoring to make it explicit in the code that you either get numerical values, or all None.

pyani_plus/private_cli.py
```diff
@@ -581,7 +586,12 @@ def fastani(  # noqa: PLR0913
                 identity,
                 orthologous_matches,
                 fragments,
-            ) in method_fastani.parse_fastani_file(tmp_output, filename_to_hash)
+            ) in method_fastani.parse_fastani_file(
+                tmp_output,
+                filename_to_hash,
+                # This is used to infer failed alignments:
+                expected_pairs={(_, subject_hash) for _ in query_hashes},
```
peterjc, line 593: Here with fastANI, as we are computing one column, the number of expected pairs equals the number of genomes, *n*. That should scale OK.

pyani_plus/private_cli.py
```diff
@@ -1102,8 +1116,15 @@ def log_branchwater(
...
-                manysearch, filename_to_hash
+            ) in method_sourmash.parse_sourmash_manysearch_csv(
+                manysearch,
+                filename_to_hash,
+                # This is used to infer failed alignments:
+                expected_pairs={
+                    (q, s)
+                    for q in filename_to_hash.values()
+                    for s in filename_to_hash.values()
+                },
```
peterjc, line 1127: Unlike fastANI were we are computing one column, here the number of expected pairs equals the number of genomes squared. That is a slight concern, but we will be refactoring the branchwater code to work on chunks rather than doing all the comparisons at once.

tests/test_coverage.py
```diff
@@ -0,0 +1,182 @@
...
+import pytest
+
+from pyani_plus import db_orm, public_cli
+
+
+def do_comparison(
+    capsys: pytest.CaptureFixture[str],
+    fasta_dir: Path,
+    method: Callable,
+    **kwargs: float,
```
peterjc, line 42: Strangely although ``**kwargs`` refers to an optional dictionary of keyword arguments, for typing only the possible values are described. In our case this is currently ``int | float`` which is recorded as ``float``.

tests/test_coverage.py
```diff
@@ -0,0 +1,181 @@
...
+    #       [null, 100%, 100%]] for query_cov
+    #
+    # Using the contig lengths, small is 7582 bp, large is 18001 bp, gives
+    # the expected coverages values vs the combined file of 29.6% and 70.4%.
+    #
+    # Comparing the JSON serialised dataframe directly from the DB
+
+    run = do_comparison(capsys, seq_dir, public_cli.anim)
+    assert run.df_identity == (
+        "{"
```
peterjc, line 115: The ``{`` is a string on its own because it messes up as part of an f-string. There could be a better solution than this?

kiepczi (approved):
Merging now!

PR #216, opened 2024-12-14
Commit ANIm, dnadiff & ANIb by column not one-by-one.

peterjc:
Previously the ANIm, dnadiff & ANIb methods were computed pairwise (one job each, meaning N² jobs for N genomes). They were then updated to compute one column per job (so just N jobs for N genomes), but this was essentially in a for-loop and each pairwise computation was still logged to the DB after computation. This was working fine with the local runner for 100s of genomes *with the DB on a local disk*.

Testing on the cluster, even with the local runner, we saw locking problems with just 40 workers *because the DB was on a network disk*. Increasing the SQLite timeout from 10s to 30s (default 5s) helped, but did not solve it.

As per https://sqlite.org/useovernet.html we need to avoid rival threads writing to the DB at the same time. To that end, this PR changes those methods to do the DB logging once per column (like the fastANI method). This means we do only N large commits, not N² individual commits, and has allowed runs with 400 worker threads to complete on the cluster.

There is scope for improvement, e.g. writing partial progress might be desirable for larger datasets. Also, if a job is interrupted, it would be nice to try to log all the cached results thus far. And we may still need to extend the timeout and/or add a retry.

One minor downside is this does mean the progress bar will sit at zero until the first task completes and then it will jump to N / N² complete (just like the fastANI implementation).

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [ ] Request a code review

### DIFF
--- a/pyani_plus/private_cli.py
+++ b/pyani_plus/private_cli.py
@@ -630,6 +630,7 @@ def anim(  # noqa: PLR0913
     )
     subject_stem = Path(hash_to_filename[subject_hash]).stem
 
+    db_entries = []
     for query_hash in query_hashes:
         query_length = (
             session.query(db_orm.Genome)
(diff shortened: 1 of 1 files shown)


kiepczi (approved):
Looks good! Thanks @peterjc. Merge now?

PR #222, opened 2024-12-20
Accept gzipped FASTA for input

peterjc:
~~This is a large step towards #124, although initially only sourmash/branchwater and fastANI work.~~

~~Additional code will be required for ANIm, dnadiff and ANIb - and perhaps switching to a different test case there for speed.~~

~~Right now this is using the 4 bacteria (now gzipped), where sourmash/fastANI need just a couple of seconds to run.~~

Using the 4 bacteria (now gzipped) for sourmash/fastANI (faster methods which already support GZIP), and a gzipped version of the 3 viruses for ANIb/ANIm/dnadiff (slower and require local decompression of GZIP).

## Type of change

## Action Checklist

- [x] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Work on this branch
- [x] Commit branch
- [ ] Request a code review


pyani_plus/db_orm.py
```diff
@@ -654,11 +655,20 @@ def db_genome(
...
+                length += len(seq)
+                if description is None:
+                    description = title  # Just use first entry
+    except gzip.BadGzipFile:
+        with Path(fasta_filename).open() as handle:
+            for title, seq in SimpleFastaParser(handle):
+                length += len(seq)
+                if description is None:
+                    description = title  # Just use first entry
+
```
peterjc, line 671: This try/except style for detecting gzip is used often enough that a helper function might be worthwhile, ideally usable as a context manager?

tests/fixtures/bacterial_example/intermediates/fastANI/all_vs_NC_002696.fna.fastani
```diff
@@ -0,0 +1,4 @@
+../fixtures/bacterial_example/NC_002696.fna.gz	../fixtures/bacterial_example/NC_002696.fna.gz	100	1338	1338
+../fixtures/bacterial_example/NC_010338.fna.gz	../fixtures/bacterial_example/NC_002696.fna.gz	82.9124	877	1825
+../fixtures/bacterial_example/NC_011916.fna.gz	../fixtures/bacterial_example/NC_002696.fna.gz	99.9946	1337	1347
+../fixtures/bacterial_example/NC_014100.fna.gz	../fixtures/bacterial_example/NC_002696.fna.gz	86.1137	998	1551
```
peterjc, line 4: I noticed there the FASTA extension has been included in the filename - a quirk of the ``pathlib.Path`` and how it handles ``Path("example.fasta.gz").stem`` as ``example.fasta``

Not a big problem for what are essentially just internal temp files, but a quirk we might want to address?

peterjc:
Currently this normalised the bacterial example back to all using ``.fna`` but in hindsight it makes sense to keep the assorted filenames in order to have tests with ``.fasta.gz`` etc too.

widdowquinn:
If we can have this work for ANIb/ANIm too, we could merge this all in one go :D

peterjc:
Rebased to use assorted ``*.f*.gz`` names matching the ``*.f*`` naming on the main branch.

peterjc:
Rebased to handle the double-suffix stem (adds a helper function).

Also adds gzip support to ANIb.

pyani_plus/db_orm.py
peterjc: Elsewhere it has been easier to switch based on the extension, so perhaps this should do that too for consistency, or do both here for sanity checking as this is part of our pipeline input routine?

pyani_plus/private_cli.py
```diff
@@ -766,9 +768,21 @@ def anib(  # noqa: PLR0913
...
 
+    if hash_to_filename[subject_hash].endswith(".gz"):
+        # We must decompress the FASTA file for makeblastdb
+        subject_fasta = tmp_dir / hash_to_filename[subject_hash][:-3]
+        subject_stem = subject_fasta.stem
+        with (
+            gzip.open(fasta_dir / hash_to_filename[subject_hash], "rb") as f_in,
+            subject_fasta.open("wb") as f_out,
+        ):
+            shutil.copyfileobj(f_in, f_out)
```
peterjc, line 781: This use of ``shutil.copyfileobj`` is taken from the https://docs.python.org/3/library/gzip.html documentation.

peterjc:
Confirmed by experimenting on macOS that ``makeblastdb`` from BLAST+ 2.16.0 will not accept ``/dev/stdin`` or a named pipe for the FASTA input. However, it does accept ``-in -`` (minus sign, also the default) to mean stdin. So we can in principle uncompress a GZIP'd FASTA file to it on the fly.

On the otherhand, ``nucmer`` from mummer 3.23 will not take either FASTA file in this way.

widdowquinn:
Is it maybe (not as a priority…) worth having a user option to be aggressive about deleting intermediate files in temporary directories, to avoid issues with caps on number of files (as with ARCHIE-West)? We haven't hit it ourselves, so it may not be a practical problem?

PR #221, opened 2024-12-20
Log ANIm, dnadiff, and ANIb in batches

peterjc:
In order to avoid locking errors writing to SQLite3 on a network drive we had switched to recording the comparisons for these methods a whole column at a time (#216) - and this has been tested with up to 400 workers on 400 genomes.

However, this change means the main application shows zero progress until the first column completes - which is not a good user experience. Quoting myself on #216:

> One minor downside is this does mean the progress bar will sit at zero until the first task completes and then it will jump to N / N² complete  (just like the fastANI implementation).

We therefore want to log DB output occasionally, in order to have the main thread progress bar update, but not too often, and not by multiple workers all at once.

Assuming all the pairwise comparisons take the same time, the design here would mean at each round a different column worker would log its progress. This would mean regular updates until all the columns complete, although at any one moment in addition to what has been logged and shown on the progress bar, there will be lots of completed calculations cached and waiting to be logged.

Note the progress bar progression will be non-linear with time, updated relatively slowly during the run (lagging the actual computational progress), and seemingly speed up and show rapid progress near the end as the columns complete.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [x] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
[...]

PR #225, opened 2024-12-29
Log pairwise computations on interrupt

peterjc:
Builds on #221.

The ANIm, dnadiff, and ANIb results are computed pairwise, but logged in batches to avoid database locking. This means at any moment there can be multiple computed results cached and waiting to be committed to the database.

Currently if the worker job were to be interrupted or terminated (e.g. via ctrl+c, ``kill``, or SLURM ``scancel``), those pending results are lost.

This change will catch the ctrl+c, SIGINT, or SIGTERM signals, and attempt to log any pending results. All the workers can be interrupted at the same time (e.g. master job interrupted), which means we may then get database locking and a failure to save everything - but we'd still save some of the computations.

This is working nicely in manual testing including ctrl+c with the local executor, and using ``kill`` to send SIGINT or SIGTERM (e.g. from a cluster node), but not via SLURM ``scancel``. This https://dhruveshp.com/blog/2021/signal-propagation-on-slurm/ hints as a possible explanation - the signal gets lost in the bash wrapper script. That may require a tweak to how we are using snakemake, or perhaps a fix in snakemake. I speculate that because snakemake tracks progress by generated files, they have no concept of a partial set of results - a cluster job either succeeds or fails, and thus this may not have been important to them.

I have not come up with a good idea for automated testing of this functionality. Probably we have to do something like https://github.com/snakemake/snakemake/blob/v8.26.0/tests/common.py#L286 (run a process in a child thread, wait some seconds, send signal).

## Type of change

## Action Checklist

- [x] Work on this branch
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Commit branch
- [x] Request a code review

### DIFF
--- a/pyani_plus/private_cli.py
+++ b/pyani_plus/private_cli.py
@@ -35,6 +35,7 @@ from typing import Annotated
 import typer
 from rich.progress import Progress
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import Session
 
 from pyani_plus import PROGRESS_BAR_COLUMNS, db_orm, tools
@@ -443,10 +444,12 @@ def compute_column(  # noqa: C901, PLR0912
         del column
 
     # What comparisons are needed?
-    query_hashes: set[str] = set(hash_to_filename).difference(
-        comp.query_hash
-        for comp in run.comparisons().where(
-            db_orm.Comparison.subject_hash == subject_hash
+    query_hashes: list[str] = sorted(
+        set(hash_to_filename).difference(
+            comp.query_hash
+            for comp in run.comparisons().where(
+                db_orm.Comparison.subject_hash == subject_hash
+            )
         )
     )
 
@@ -509,7 +512,7 @@ def fastani(  # noqa: PLR0913
     fasta_dir: Path,
     hash_to_filename: dict[str, Path],
     filename_to_hash: dict[str, str],
-    query_hashes: set[str],
+    query_hashes: list[str],
     subject_hash: str,
     *,
     quiet: bool = False,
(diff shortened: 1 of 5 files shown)


peterjc:
Cross reference https://github.com/snakemake/snakemake/issues/3241

widdowquinn:
I think this is a great thing to have for local jobs. I don't know how much work is needed to make it work the same way on distributed SLURM jobs. I'm inclined to merge for having the local capability, but it would be nice to have it work on distributed jobs, too… how much work does that need?

If it would take to long, we could merge as-is, and include the caveat in the documentation that it might not work as expected on SLURM/SnakeMake.

peterjc:
The test will need tweaking, we need to wait a shorter period than on my laptop:

``AssertionError: Expected partial progress, not 3/3 and return 0``

Probably it needs to use a larger test case in order have a reliable timing (e.g. one column of a 10 genome example)?

peterjc:
The test approach is inspired by how snakemake tests with sigint:

https://github.com/snakemake/snakemake/blob/v8.26.0/tests/common.py#L292

peterjc:
Given the timing factor, building a test case suitable for GitHub Actions AND local full test (e.g. ``make test`` in parallel) AND local test of just ``test_private_cli.py`` (single threaded) was a slight challenge. I just make the test larger so that even on a faster machine it shouldn't finish in the window before the SIGINT is sent.

Also, we can use the same basic technique to test https://github.com/pyani-plus/pyani-plus/blob/v0.0.1a1.dev2/pyani_plus/workflows/__init__.py#L203

peterjc:
Refreshed with private-cli interrupt tests which seem to work nicely on my local machine (including with parallel testing via ``make test`` or directly) and on GitHub actions.

I could not make a reliable public-cli interrupt test, but on the bright side could remove an unused error handler. I have also included assorted ``pragma: no cover`` annotations to exclude that and a few other corner cases where testing is impractical.

That leaves just 2 lines for catching a ``blastn`` vs ``makeblastdb`` version mismatch as the only code missing tests or excluded - and I can see how to test.

peterjc:
I'm merging this now as it has conflicts with the GZIP work which is hampering my testing.

PR #231, opened 2025-01-10
User shorter filenames for snakemake workflow tests

peterjc:
Simple PR (which can easily be rebased if #220 is merged first) to use shorter filenames, meaning the pytest output works better on narrower screens.

PR #233, opened 2025-01-12
Remove methods folder, leaving flat source code tree

peterjc:
Something for discussion:

* Good:
  * [Zen of Python](https://peps.python.org/pep-0020/): "Flat is better than nested." 
  * Shorter names for pytest output (cross reference #231)
* Bad:
  * This was a logical grouping of modules
  * One could argue the verbose pytest doctest output is a bug


widdowquinn:
I don't think flattening hurts the codebase, but I do like the clarity of the hierarchy.

This issue isn't realy user-facing or functional, so if we can live with the `pytest` long strings I think we maybe ought to, meaning we shorten the file name, rather than losing the hierarchy, to achieve the same goal.

widdowquinn:
However, if the issue is that we need to avoid confusion between, e.g. a method `sourmash()` and a module `sourmash`, but we want to avoid the redundancy of `methods/method_sourmash.py` - then flattening is probably the right way to go.

peterjc:
Alternative plan would be to drop the prefix in the filename but adjust the imports to compensate. Thus old:

```python
from pyani_plus.methods import method_sourmash
```

becomes:

```python
from pyani_plus.methods import sourmash as method_sourmash
```

This avoids the ambiguity in ``sourmash`` (which we use as a function name in the public API for example).

Issue #228, opened 2025-01-06
add `pyani classify` module


kiepczi:
We intend to include the `pyani classify` command in `pyani-plus`. The purpose of `pyani classify` is to identify cliques (k-complete, or fully-connected sub-graphs) within a set of genomes, generating potentially meaningful groups of genomes.

To identify groups of genomes, the following steps can be taken:

1. Create a complete graph: Construct a graph where each node represents a genome, and edges correspond to pairwise comparisons (e.g., %ID or %COV).
2. Apply an initial threshold: Remove edges based on a specified threshold (e.g., 50% genome coverage or 95% identity). This step will split the complete graph into clusters.
3. Analyse cluster properties: Examine the resulting clusters to determine their properties (e.g., cliques versus non-cliques) and report the number of cliques.
4. Test other thresholds: When the first threshold is tested, we can apply next thresholds by removing edges with the lowest value/score and repeat until all nodes are siglentons. 

Desired output is a list of cliques together with the (upper and) lower threshold which they appear at, and ideally a plot visualising this.

kiepczi:
Working on this on branch `classify`.

kiepczi:
A quick update on this issue: there has been some progress in #243. However, there are still a few things I need to look into:

- What happens if there is no alignment found between some genomes? Does the code break or falsely retain edges between the genomes? Currently, we remove edges with the lowest threshold assigned to them, but if such edges do not have a value, are the edges kept or not?
- What happens if there are missing values in the database? For example, what if all identity values are found, but no coverage values are present? (I run into `RecursionError` while testing and this is what is likley to happen in this case).
- Currently, the output in the TSV file reports members based on stem files. We might want to make this optional and allow custom labels, filenames, or hashes instead.

widdowquinn:
>     * What happens if there is no alignment found between some genomes? Does the code break or falsely retain edges between the genomes? Currently, we remove edges with the lowest threshold assigned to them, but if such edges do not have a value, are the edges kept or not?

If there is no alignment between two genomes, there is no edge between the nodes representing those genomes.

>     * What happens if there are missing values in the database? For example, what if all identity values are found, but no coverage values are present? (I run into `RecursionError` while testing and this is what is likley to happen in this case).

I think this should raise an error, if coverage is being used in the criteria for finding cliques (e.g. in the initial winnowing of edges).

[...]

PR #243, opened 2025-01-15
Add pyani-plus classify command to find cliques

kiepczi:
This addresses #228, as discussed earlier this week at the meeting.

Currently, the classify method:
- Constructs a graph where nodes represent genomes and edges correspond to pairwise comparisons. Since ANI methods may not be symmetric (e.g., different values reported for genome A vs. genome B and genome B vs. genome A), users can specify which values to use (e.g., min, max, or mean). The default settings are min for coverage and mean for identity.
- Classifies genomes into cliques recursively by removing the lowest edges and analysing the graph at each step to identify cliques, recording them when found.
- Outputs two TSV files (one for coverage and one for identity) containing information about clique members, the number of members, minimum coverage, and minimum identity within each clique.

Additionally, classify has been added to the public CLI.

## Type of change

## Action Checklist

- [X] Work on this branch
- [X] Commit branch
- [X] Request a code review


tests/test_public_cli.py
```diff
@@ -1400,3 +1400,71 @@ def test_plot_bad_nulls(
...
+        public_cli.classify(database=tmp_db, outdir=tmp_dir)
+    tmp_db.unlink()
+
+
+def test_classify_db(
+    capsys: pytest.CaptureFixture[str], input_genomes_tiny: Path, tmp_path: Path
+) -> None:
+    """Check classify working example."""
+    tmp_dir = Path(tmp_path)
+    db = input_genomes_tiny / "viral_database.db"
```
peterjc, line 1464: We should avoid having binary blobs like SQLite databases under version control. See some of the other tests which create a DB in code using mock values - or just call something very fast like sourmash or fastANI on the three viruses as part of this test?

pyani_plus/public_cli.py
```diff
@@ -42,12 +43,15 @@
 from sqlalchemy.orm import Session
 
 from pyani_plus import PROGRESS_BAR_COLUMNS, db_orm, tools
+from pyani_plus import classify as method_classify
```
peterjc, line 46: This isn't an ANI method, so I don't think it needs the ``method_`` prefix.

pyani_plus/public_cli.py
```diff
@@ -870,5 +874,72 @@ def plot_run(  # noqa: C901
...
+        outdir / f"{method}_coverage_classify.tsv", sep="\t", index=False
+    )
+
+    id_cliques_df = pd.DataFrame(id_cliques_info)
+    id_cliques_df["members"] = id_cliques_df["members"].apply(lambda x: ",".join(x))
+    id_cliques_df.to_csv(
+        outdir / f"{method}_identity_classify.tsv", sep="\t", index=False
+    )
+
+    print(f"Wrote classify output to {outdir}")
```
peterjc, line 939: I think most of this function should be moved into the ``classify.py`` file.

tests/test_public_cli.py
```diff
@@ -1400,3 +1400,71 @@ def test_plot_bad_nulls(
...
+        SystemExit, match="ERROR: Output directory /does/not/exist does not exist"
+    ):
+        public_cli.classify(database=":memory:", outdir=Path("/does/not/exist/"))
+
+    tmp_db = tmp_dir / "export.sqlite"
+    session = db_orm.connect_to_db(tmp_db)
+    with pytest.raises(SystemExit, match="ERROR: Database contains no runs."):
+        public_cli.classify(database=tmp_db, outdir=tmp_dir)
+
+    config = db_orm.db_configuration(
```
peterjc, line 1424: I don't think you need the rest of this test - I believe all those errors are via helper functions which are already tested. You can try removing the test and confirming the coverage is unchanged?

tests/test_public_cli.py
kiepczi: I tried to copy the approach you used in tes_resume_partial_fastani, where you set up a database in code with mock values, but I ran into a FAILED tests/test_public_cli.py::test_partial_run - RecursionError: maximum recursion depth exceeded error. I'll take another look—maybe I missed something...

pyani_plus/public_cli.py
kiepczi: That's fair. I had to import this as something because there's a conflict with the public CLI:
```python
@app.command()
def classify(
    database: REQ_ARG_TYPE_DATABASE,
    outdir: REQ_ARG_TYPE_OUTDIR,
    covearge_edges: OPT_ARG_TYPE_COVERAGE_EDGES = "min",  # Default to "min"
    identity_edges: OPT_ARG_TYPE_IDENTITY_EDGES = "mean",  # Default to "mean"
    run_id: OPT_ARG_TYPE_RUN_ID = None,
) -> int:
```

The `method_` prefix was the first thing that came to mind. Other options I thought of are `classify_cliques` or `classify_genomes`. Do you have any other recommendations?

tests/test_public_cli.py
peterjc: Recursion? That was unexpected from setting up a mock DB.

kiepczi: Sorry for not being clear. Setting up a mock database didn’t cause the error. The error occurred when I tried to run `classify` on that mock database. I’m assuming the recursion error happened because I use a recursive function to find cliques? I will have another look and see what happens... If I replicate the error, I will try to share it here...

kiepczi: I don't know how I missed it yesterday, but the `RecursionError` is raised when there is information missing in the DB. I accideltly only added values for identity and not the covearge 😥... I will need to optimise `pyani classify` to deal with diffrent scenarios, but it might be best to do it on a separate branch...

pyani_plus/public_cli.py
kiepczi: I changed this to `classify_cliques`.

pyani_plus/public_cli.py
kiepczi: Done!

tests/test_public_cli.py
kiepczi: The coverage hasn't change after removing this, which is good, but there are few things that I can't seem to hit with testing yet..

pyani_plus/public_cli_args.py
```diff
@@ -160,3 +160,18 @@
...
+    ),
+]
+
+OPT_ARG_TYPE_IDENTITY_EDGES = Annotated[
+    str,
+    typer.Option(
+        help="How to resolve asymmetrical ANI identity results for edges in the graph (min, max or mean).",
+        rich_help_panel="Method parameters",
+    ),
+]
```
peterjc, line 177: _If_ these are only likely to be used for the one public CLI function, perhaps they can live directly in the function definition?

requirements.txt
```diff
@@ -16,3 +16,4 @@ rich
 seaborn>=0.13.2
 sqlalchemy>=2.0
 typer>=0.12
+networkx
```
peterjc, line 19: I would put at least a minimum of the current major.minor version - I know from experience that there have been breaking changes in the networkx API.

pyani_plus/public_cli.py
```diff
@@ -870,5 +873,60 @@ def plot_run(  # noqa: C901
...
+        print(f"INFO: Exporting run-id {run_id}")
+
+    mapping = {_.genome_hash: Path(_.fasta_filename).stem for _ in run.fasta_hashes}
+
+    identity = run.identities
+    if identity is not None:
+        identity.rename(index=mapping, columns=mapping, inplace=True)  # noqa: PD002
+    cov = run.cov_query
+    if cov is not None:
+        cov.rename(index=mapping, columns=mapping, inplace=True)  # noqa: PD002
```
peterjc, line 906: I suggest dropping your custom renaming and use the same code as export-runs and plot (on both matrices). You can add the same ``--label`` command line option.

tests/test_public_cli.py
```diff
@@ -1400,3 +1400,68 @@ def test_plot_bad_nulls(
...
+        session,
+        config,
+        cmdline="pyani guessing ...",
+        fasta_directory=input_genomes_tiny,
+        status="Complete",
+        name="Test classify when all data present",
+        fasta_to_hash=fasta_to_hash,
+    )
+
+    public_cli.classify(database=tmp_db, run_id=1, outdir=tmp_dir)
```
peterjc, line 1462: If you leave out the run ID this will test the code picking the latest run (something noted on the test coverage as missing).

peterjc:
(PR re-titled, the typo was bugging me - sorry)

pyani_plus/public_cli.py
```diff
@@ -870,5 +859,82 @@ def plot_run(  # noqa: C901
...
+    if cov is None:
+        msg = f"ERROR: Could not load run {method} matrix"  # pragma: no cover
+        sys.exit(msg)  # pragma: no cover
+
+    try:
+        identity = run.relabelled_matrix(identity, label)
+        cov = run.relabelled_matrix(cov, label)
+    except ValueError as err:
+        msg = f"ERROR: {err}"
+        sys.exit(msg)
```
peterjc, line 915: This exception isn't currently covered in the tests. You can just add a call to your new public API in the ``test_export_duplicate_stem`` function.

widdowquinn:
Do we want a test case where there is a valid comparison in one direction, but not the other, so the graph construction attempts to create an edge with one valid value, and one `NULL`?

Or should we only allow clique analysis to proceed if the matrix is complete, and throw a graceful error/other message if it is not?

widdowquinn:
This looks pretty good but can we have a more complex test case where we're finding known subgraph cliques and have to use recursion to do so?

widdowquinn:
Not for this branch, but once this branch is merged, we could identify the clique identity boundaries within the recursive algorithm.

If a subgraph is a clique, then removing the lowest-value edge will break the clique. That's an upper threshold on the _range_ of the clique (i.e. the identities across which it exists).

The lower threshold would be the weight on the edge that was removed immediately before the clique formed. This could be represented as a new attribute to `find_cliques()` (e.g. `last_edge_weight`) so that when a clique is found, we can record the range of identities over which it exists.

Cliques with larger ranges might be more stable (and thus more biologically interesting) than those with shorter ranges. If there's a range of values over which many cliques are valid, that might represent something biologically interesting too.

This also lends itself to interesting options for visualisation.

peterjc:
Can you try building that 6-node example as a test case? Where it breaks up into two 3-node cliques? You can give the nodes/genomes arbitrary names like A to F.

peterjc:
Ah - I think you've done something like that on #255.

Shall we close this (#243)?

PR #249, opened 2025-01-17
Default heatmap colour schemes from legacy pyANI v0.3

peterjc:
Builds on #232 by:
* refactoring to move the plotting into its own module
* adding the default red/blue based colour schemes, figure size, and line width settings from legacy pyANI v0.3 (as applicable to the three heatmaps currently being drawn).

Note the lazy import to avoid importing ``matplotlib`` and ``seaborn`` etc for the other commands.

Also note the heatmaps remain difficult to use beyond a certain number of genomes.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [x] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #255, opened 2025-01-23
add pyani-classify command to identify cliques based on ANI results

kiepczi:
This PR is in favor of #243.
The way cliques are identified has been tweaked a bit. To keep things simple and avoid complicating the history while implementing `pyani-plus classify`, I decided it would be easier to start fresh with a new branch.

In this PR, I’m suggesting the following steps for identifying cliques:

1. Building the Initial Graph:
- We create an initial graph where nodes are genomes and edges are pairwise comparisons. Since some ANI methods might not be symmetrical, users can choose which value to use (e.g., `min`, `max`, or `mean`). By default, we use `min` for coverage and `mean` for identity.
- If no alignment is found between a pair of genomes, we don’t add an edge.
- Any edges where the genome coverage is below a certain threshold (50% by default) are removed. This keeps the number of edges smaller, which means fewer checks for cliques and faster processing. But this threshold can be changed if users need a diffrent threshold else.

2. Finding Potential Cliques:
- We check the graph for potential cliques. At this point, the results might split into multiple connected components, with some being cliques. We’ll want to catch this before removing any edges.

3. Removing Lowest Identity Edges:
- We remove edges with the lowest identity one by one, checking for cliques as we go. We keep doing this until all the nodes are singletons.

4. Saving the Results:
- The analysis is saved to TSV files, showing information about each potential clique, including its members, the number of members, minimum coverage, and minimum identity.

## Action Checklist

- [X] Work on this branch
  - [X] corresponding changes to documentation have been made
- [X] Commit branch
- [X] Request a code review


pyani_plus/public_cli.py
```diff
@@ -1004,5 +993,86 @@ def plot_run(  # noqa: C901
...
+
+    # Construct the graph with the correct functions
+    complete_graph = classify.construct_graph(
+        cov, identity, covearge_agg_func, identity_agg_func, cov_min
+    )
+    # Finding cliques
+    if len(list(nx.connected_components(complete_graph))) != 1:
+        initial_cliques = classify.find_initial_cliques(complete_graph)
+    else:
+        initial_cliques = []
```
peterjc, line 1064: Probably a trivial fully connected graph would cover this branch? May even a single node?

pyani_plus/classify.py
```diff
@@ -0,0 +1,194 @@
...
+        graph.remove_edge(edge_to_remove[0], edge_to_remove[1])
+
+        connected_components = list(nx.connected_components(graph))
+        if len(connected_components) > 1:
+            for component in connected_components:
+                subgraph = graph.subgraph(component).copy()
+                cliques.extend(find_cliques_recursively(subgraph, progress, task))
+            return cliques
+
+    return cliques
```
peterjc, line 165: Another corner case with no coverage, perhaps testing with a single node graph - or a two node graph with no edges?

widdowquinn:
A database of 100 singletons… 
[random.db.zip](https://github.com/user-attachments/files/18557510/random.db.zip)

widdowquinn:
We need to expand the test cases to ensure that the classification is working as we expect (e.g. inputs with only singletons, graphs that start off as disconnected subgraphs, graphs that become disconnected subgraphs, etc.) - we only need small examples, really so that we're focusing in on correct operation in each case.

widdowquinn:
A database of synthetic sequence examples, where we don't know the correct answers ;) [synthcomp.db.zip](https://github.com/user-attachments/files/18557529/synthcomp.db.zip)

peterjc:
Personally I think the corner case of a single genome in the database, with only a single self-vs-self comparison (which will _typically_ be 100% identity, 100% coverage - not not true in all methods and we have test cases for that), is still valid, and should not give an error. Rather it should give a single clique.

kiepczi:
> Personally I think the corner case of a single genome in the database, with only a single self-vs-self comparison (which will _typically_ be 100% identity, 100% coverage - not not true in all methods and we have test cases for that), is still valid, and should not give an error. Rather it should give a single clique.

That's a fair point. I'm happy to revert this to what it was before, but would reporting a single clique make sense? In classify, each genome is represented as a node, and self-to-self comparisons are not taken into account at all (e.g., these are not presented as self-loops).

The question is: what would be more informative—returning a .tsv file with a single row containing just one member and no other information, or raising an error like this:
ERROR: Run 1 has 1 comparison across 1 genome. No edges to remove.

It might come down to personal preference, but I would rather have an error that flags the issue when analysing a problematic run. After all, what is the point of running classify on a single genome? 🤔

peterjc:
I agree how to handle the single genome corner case is largely personal preference (where there is a logical/reasonable behaviour). For the plot command however, our choice may be forced by errors/warnings from the underlying libraries (e.g. distribution plots with only a single score like the everything-is-awesome 100% coverage plot discussed yesterday on #249).

kiepczi:
So, do you think that reproting a single clique in this case would be the best option? I'm getting a little bit confused. 

What are your thought on this @widdowquinn?

kiepczi:
Thinking about this, we can compromise on this one, and still report this as a single clique, but in the case where the run has a single comparison across 1 genome, to perhaps use a print statment saying something like this: `Run 1 has 1 comparisons across 1 genomes. Is this correct?`

widdowquinn:
> That's a fair point. I'm happy to revert this to what it was before, but would reporting a single clique make sense? In classify, each genome is represented as a node, and self-to-self comparisons are not taken into account at all (e.g., these are not presented as self-loops).

The end state of the recursive clique-finding algorithm is that the final "clique" (the k-complete graph for k=1, i.e. a clique of size 1) is a singleton.

The described behaviour: a single genome in the database run gives a single clique with that genome as a member, is the correct behaviour.
 
> It might come down to personal preference, but I would rather have an error that flags the issue when analysing a problematic run. After all, what is the point of running classify on a single genome? 🤔

When we have logging implemented (and user warnings with that system) we would raise this as a warning, because it is a weird request. We would also - I hope - have raised a warning that that the comparison set to generate the database contained only a single genome sequence.

peterjc:
Agreed - in the short term ``sys.stderr.write("WARNING: Only one genome\n")`` or similar would do the trick.

pyani_plus/public_cli.py
kiepczi: Done!

pyani_plus/classify.py
kiepczi: Done!

tests/test_classify.py
```diff
@@ -0,0 +1,331 @@
...
+
+    clique_3 = nx.Graph()
+    clique_3.add_node("genome_2")
+
+    return [clique_1, clique_2, clique_3]
+
+
+@pytest.fixture
+def known_complex_graph() -> nx.Graph:
+    """Return nx.Graph with six nodes and known weights (coverage and identity)."""
```
peterjc, line 109: Can you describe it in a paragraph? e.g. This six node graph first splits into two cliques of three-nodes.

tests/test_classify.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+@pytest.fixture
+def two_nodes_no_edges_dataframes() -> list[pd.DataFrame]:
+    """Return dataframes for building initial graph with two nodes and no edges."""
+    # Define the genomes
+    genomes = ["genome_1", "genome_2"]
+
+    # Create the DataFrame directly with initial values
+    # When building graph with `onstruct_graph` no edges should be added if coverage is <0.50
```
peterjc, line 73: onstruct_graph -> construct_graph

pyani_plus/public_cli.py
```diff
@@ -939,5 +941,97 @@ def plot_run(
...
+def cli_classify(  # noqa: PLR0913
+    database: REQ_ARG_TYPE_DATABASE,
+    outdir: REQ_ARG_TYPE_OUTDIR,
+    coverage_edges: Annotated[
+        str,
+        typer.Option(
+            help="How to resolve asymmetrical ANI identity results for edges in the graph (min, max or mean).",
+            rich_help_panel="Method parameters",
+        ),
+    ] = "min",  # Default to "min"
```
peterjc, line 954: No need for this comment (same below)

pyani_plus/classify.py
```diff
@@ -0,0 +1,193 @@
...
+}
+
+MIN_COVERAGE = 0.50
+
+
+class CliqueInfo(NamedTuple):
+    """Graph structure summary."""
+
+    members: list
+    n_nodes: int
```
peterjc, line 49: I see no reason to record number of nodes explicitly when you can use ``len(n.members)`` rather than ``n.n_nodes`` which prevents this ever being inconsistent.

[If this was an expensive calculation then I could be convinced a cached property was OK]

kiepczi: That's fair. I think the information stored here will change, as we talked earlier this week about including details like the previous threshold that led to the formation of the clique. I was thinking of making that change in the next PR. Should I leave out dropping n_nodes for now, or should I do it on this branch?

peterjc: Up to you - unless Leighton has time to look at it first, I don't think we'll merge this until Monday, so this PR?

widdowquinn:
Can we please round identities to 7dp (otherwise Excel treats them as characters which is a pain in the arse), and move the clique members to the end column, for aesthetics?

Also there's a duplication issue with the `strep.db. data that might need a tweak:

```
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/lpritc/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/lpritc/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>
<!--table
	{mso-displayed-decimal-separator:"\.";
	mso-displayed-thousand-separator:"\,";}
@page
	{margin:.75in .7in .75in .7in;
	mso-header-margin:.3in;
	mso-footer-margin:.3in;}
tr
	{mso-height-source:auto;}
col
	{mso-width-source:auto;}
br
	{mso-data-placement:same-cell;}
td
	{padding-top:1px;
	padding-right:1px;
	padding-left:1px;
	mso-ignore:padding;
	color:black;
	font-size:12.0pt;
	font-weight:400;
	font-style:normal;
	text-decoration:none;
	font-family:"Aptos Narrow", sans-serif;
	mso-font-charset:0;
	mso-number-format:General;
	text-align:general;
	vertical-align:bottom;
	border:none;
	mso-background-source:auto;
	mso-pattern:auto;
	mso-protection:locked visible;
	white-space:nowrap;
	mso-rotate:0;}
.xl63
	{mso-number-format:"0\.0000000";}
-->
GCF_003610995.1_ASM361099v1_genomic,GCF_010548165.1_ASM1054816v1_genomic,GCF_016906245.1_ASM1690624v1_genomic,GCF_004000405.1_ASM400040v1_genomic,GCF_001434355.1_ASM143435v1_genomic,GCF_001905905.1_ASM190590v1_genomic,GCF_001746305.1_ASM174630v1_genomic,GCF_009863145.1_ASM986314v1_genomic,GCF_018619185.1_ASM1861918v1_genomic,GCF_000010605.1_ASM1060v1_genomic,GCF_000718455.1_ASM71845v1_genomic,GCF_002094995.1_ASM209499v1_genomic,GCF_016901035.1_ASM1690103v1_genomic,GCF_002910905.1_ASM291090v1_genomic,GCF_000261345.2_ASM26134v2_genomic,GCF_002551175.1_ASM255117v1_genomic,GCF_013364315.1_ASM1336431v1_genomic,GCF_002551145.1_ASM255114v1_genomic,GCF_002382885.1_WZ.A104_genomic | 19 | 0.54034186 | 0.8864437
-- | -- | -- | --
GCF_003610995.1_ASM361099v1_genomic,GCF_010548165.1_ASM1054816v1_genomic,GCF_016906245.1_ASM1690624v1_genomic,GCF_004000405.1_ASM400040v1_genomic,GCF_001434355.1_ASM143435v1_genomic,GCF_001905905.1_ASM190590v1_genomic,GCF_001746305.1_ASM174630v1_genomic,GCF_009863145.1_ASM986314v1_genomic,GCF_018619185.1_ASM1861918v1_genomic,GCF_000010605.1_ASM1060v1_genomic,GCF_000718455.1_ASM71845v1_genomic,GCF_002094995.1_ASM209499v1_genomic,GCF_016901035.1_ASM1690103v1_genomic,GCF_002910905.1_ASM291090v1_genomic,GCF_000261345.2_ASM26134v2_genomic,GCF_002551175.1_ASM255117v1_genomic,GCF_013364315.1_ASM1336431v1_genomic,GCF_002551145.1_ASM255114v1_genomic,GCF_002382885.1_WZ.A104_genomic | 19 | 0.54034186 | 0.8864437
```

widdowquinn:
Closes #228

Issue #242, opened 2025-01-15
Add tANI to output reports for methods, where appropriate.


widdowquinn:
## Summary:

tANI is an alternative measure that incorporates both coverage and identity, which can be calculated and reported for more than one method currently on offer in `pyani-plus`.  In [doi:10.1093/sysbio/syab060](https://doi.org/10.1093/sysbio/syab060) it is reported to be approximately linear with phylogenetic distance, avoiding saturation effects (which are stated, but I think not shown, in that paper to be stronger for ANIm than ANIb), and producing distance topologies that are comparable to MLSA.

## Description:

tANI (total Average Nucleotide Identity) is a variant on the Hadamard matrix we already report (i.e. coverage * identity), where

tANI = -ln(coverage * identity)

This may require only a small addition to `pyani-plus` for, e.g. ANIm and ANIb.

peterjc:
I wonder if both Hadamard and tANI need to be stored in the database, versus computed on demand from the identity and query-coverage matrices? i.e. Should we add tANI to the comparisons table, or should we remove Hadamard from the comparisons table?

I agree it makes sense to restrict these by method since in several cases we only have a proxy measure for coverage, putting the computed Hadamard or tANI on shaky ground.

widdowquinn:
I don't know whether to advise extreme user caution/scepticism when using tANI for anything other than ANIm/ANIb-like methods, where coverage is intuitively sensible - or just not perform the calculation.

Being thorough for a comparison of what is, and is not, useful from each method we'd want access to those values - so I could be swayed to calculating it regardless, even though it's probably only meaningful for a subset of methods.

peterjc:
Agreed. It would be simpler to always compute tANI and Hadamard if we have both identity and coverage values 😉

peterjc:
Any preferences on the colour scheme for plotting this? Goes from zero for perfect (100% identity, 100% coverage, gives -log(1.0) = zero) upwards to a theoretical infinity as identity and/or coverage drops towards zero. Probably makes sense to apply a threshold and thus artificial upper bound?

widdowquinn:
I hadn't thought about that, and I think you're right.

Realistically, none of the ANI methods are going to get down below about 70% identity. Coverage is a bit less constrained and could unhelpfully approach zero. Gosselin _et al._ have a threshold of 5 (arbitrary units) in their figure 1, so we could lazily follow that. It's not an unreasonable number to go for, for other reasons, either:

[...]

PR #264, opened 2025-01-30
Use same red/blue on all heatmaps

### LINKED ISSUE #263: Coverage heatmap palette - does it match the other heatmaps?
While reviewing #262 I noticed that the palettes for identity/hadamard/coverage weren't the same. We should bring coverage in line with the more muted tones of identity/hadamard.

Image

peterjc:
Quick fix to close #263.


peterjc:
I'm going to merge this now - it will make life slightly easier should we rework the colour scheme as per discussion on #242

PR #273, opened 2025-02-10
Plot run comparisons

peterjc:
Sample usage:

```console
❯ pyani-plus plot-run-comp --database fakey_sample.db --outdir /tmp/ --run-ids 1,2,3,5
INFO: Plotting 3 runs against ANIb run-id 1
INFO: Run 1 has 10000 comparisons
INFO: Plotting ANIm run 2 vs ANIb run 1
INFO: Plotting ANIm run 2 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting dnadiff run 3 vs ANIb run 1
INFO: Plotting dnadiff run 3 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting fastANI run 5 vs ANIb run 1
INFO: Plotting fastANI run 5 vs ANIb run 1, with 10000 comparisons in common
Plotting ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:02 3/3
Wrote 15 images to /tmp/ANIb_identity_1_vs_*.*
```


peterjc:
This is currently a minimal set of functionality, doing a separate plot for each comparisons (1 vs 2, 1 vs 3, 1 vs 4, ...) from reference 1, versus 2, 3, 4, ...

Could do these side by side (tiled, e.g. matplotlib `GridSpec`) as a single image.

They should all have the same x and y limits.

Would also like a difference plot (on the y-axis)

Adding the identity x=y line to the scatter.

Not keen on overplotting with different colours for each run (won't scale).

peterjc:
Sample output from the above command once again:

```console
❯ pyani-plus plot-run-comp --database fakey_sample.db --outdir /tmp/ --run-ids 1,2,3,5
INFO: Plotting 3 runs against ANIb run 1
INFO: Run 1 has 10000 comparisons
INFO: Plotting ANIm run 2 vs ANIb run 1
INFO: Plotting ANIm run 2 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting dnadiff run 3 vs ANIb run 1
INFO: Plotting dnadiff run 3 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting fastANI run 5 vs ANIb run 1
INFO: Plotting fastANI run 5 vs ANIb run 1, with 10000 comparisons in common
Plotting ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  0:00:01 3/?
Wrote 15 images to /tmp/ANIb_identity_1_vs_*.*
```

Currently three separate plots (runs 2, 3, 5 vs 1):

ANIb_identity_1_vs_2
ANIb_identity_1_vs_3
ANIb_identity_1_vs_5

This is using seaborn ``jointplot`` which internally uses ``matplotlib`` subfigures. It appears nesting that within more ``matplotlib`` subfigures isn't possible - so instead we'd have to handle *all* the subfigures and mimic some of what ``jointplot`` does. This is good in that I'd not want to repeat the marginal plot of the reference run scores each time.

widdowquinn:
This looks good as it is and provides the base functionality we need.

I know there's another branch with tweaks in prep including the shared y-axis limits for scatterplots (so visual comparisons are more informative) and tiling/separation of differences/scatterplots.

Happy to merge now - thank you!

PR #250, opened 2025-01-17
New command to import external-alignment giving ANI values

peterjc:
WIP implementing #244 from a FASTA MSA file.

TODO: For resume mode, skip already done comparisons.
TODO: Slow enough to do by column? ~7 mins for 100 genomes, each 1 million bp

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [ ] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #235, opened 2025-01-13
Lazy imports for numpy and pandas in DB ORM code

peterjc:
Prompted by ``Tests/test_interupt.py`` _sometimes_ failing locally as the private CLI was still importing pandas when the SIGINT was received.

The only complication here is the use of ``pd.DataFrame`` in the type annotation, which is now handled via an ``if TYPE_CHECKING`` condition. Sadly the initial simple version of this didn't work with ``pytest``, and I'm not really happy with the workaround.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [ ] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest-plus -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #276, opened 2025-02-11
Support KeyboardInterrupt in compute external-alignment

peterjc:
The code change is mostly indentation for the new try/except to match the other by-column computation methods.

This includes a test of an artificial large MSA (large enough that it won't have finished running the column even when no other tests are run). The interrupt delay has to be several seconds because when running multiple tests it would otherwise interrupt during the python import stages. i.e. Similar timing issues to the other interrupt tests.

Follow up to #250 

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review

### DIFF
--- a/pyani_plus/private_cli.py
+++ b/pyani_plus/private_cli.py
@@ -1414,9 +1414,9 @@ def compute_external_alignment(  # noqa: C901, PLR0912, PLR0913, PLR0915
                     )
     except KeyboardInterrupt:
         # Try to abort gracefully without wasting the work done.
-        msg = f"Interrupted, will attempt to log {len(db_entries)} completed comparisons\n"
-        sys.stderr.write(msg)
-        run.status = "Worker interrupted"
+        msg = f"Interrupted, will attempt to log {len(db_entries)} completed comparisons\n"  # pragma: no cover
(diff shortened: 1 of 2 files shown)


peterjc:
Getting this test to work reliably under GitHub Actions on macOS, Linux, and local testing single threaded and with multiple workers, is tricky.

The interruption tests all send the interruption after a fixed time which must be long enough for the private CLI to have actually started computation - and this seems to need several seconds in local testing with multiple workers. Here this was set at 4s. Then we need the test case to be big enough that it hasn't finished in 4s - the temporary mock alignment file is already 191MB (100 mock genome entries of 2 million bp).

I could try increasing the temporary mock alignment file further, but not a good idea for performance. A more elegant fix would be to scale the 4s delay dynamically...

[Also, working on cutting down the Python import overheads for the private CLI might be more worthwhile, see e.g. #235]

peterjc:
I'm not getting something like this to work reliably locally:

```python
time.sleep(5 if int(os.environ.get("PYTEST_XDIST_WORKER_COUNT", 1)) > 1 else 1)
```

I think the timings are just to finely balanced to find a one-size-fits all test.

PR #282, opened 2025-02-17
Plot run comparison scatter & differences plots separately

peterjc:
Follow up to #273 doing the scatter and difference plots as separate images, and shared y-axis for scatters.

The sub-plots are by default tiled, but you can set the number of columns - use one column and place the scatter and difference images side-by-side to recover the previous output.

### DIFF
--- a/pyani_plus/plot_run.py
+++ b/pyani_plus/plot_run.py
@@ -551,7 +551,7 @@ def plot_run_comparison(  # noqa: C901, PLR0912, PLR0913, PLR0915
         for ext in formats:
             filename = (
                 outdir
-                / f"{run.configuration.method}_{field}_{mode}_{run.run_id}_vs_others.{ext}"
+                / f"{run.configuration.method}_{field}_{run.run_id}_{mode}_vs_others.{ext}"
             )
             if ext == "tsv":
                 pass
--- a/pyani_plus/plot_run.py
+++ b/pyani_plus/plot_run.py
@@ -425,8 +425,8 @@ def plot_run_comparison(  # noqa: C901, PLR0912, PLR0913, PLR0915
         for plot_number in range(1, vs_count):
             scatter_axes[plot_number] = fig.add_subplot(
                 gs[
-                    1 + (plot_number % plots_per_col),
-                    3 * (plot_number // plots_per_col),
+                    1 + (plot_number // plots_per_row),
+                    3 * (plot_number % plots_per_row),
                 ],
                 sharex=scatter_axes[0],
                 sharey=scatter_axes[0] if mode == "scatter" else None,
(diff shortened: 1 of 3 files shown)


kiepczi:
This acts a bit weird sometimes... I run it on a fakey database with this command, where I set the columns to 3:
```bash
(pyani_plus_dev) akiepas@macstudio pyani-plus % pyani-plus plot-run-comp --database ../plotting/fakey_sample.db --outdir ../plotting --run-ids 1,2,3,4,5 --columns 3
INFO: Plotting 4 runs against ANIb run 1 which has 10000 comparisons
INFO: Plotting ANIm run 2 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting dnadiff run 3 vs ANIb run 1, with 10000 comparisons in common
INFO: Plotting branchwater run 4 vs ANIb run 1, with 9971 comparisons in common
INFO: Plotting fastANI run 5 vs ANIb run 1, with 10000 comparisons in common
Plotting scatter  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 4/4
Plotting diff     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 4/4
Wrote 8 images to ../plotting/ANIb_identity_1_vs_*.*
```
...and the plots were a bit messed up. Is there a way to fix this (see below)?
ANIb_identity_diff_1_vs_others

kiepczi:
FYI, it worked fine when the `--columns` param was set to 1,2,4,5 etc. It seems like it does not like 3 in this case

peterjc:
That behaviour does actually make sense to me - a corner case I didn't catch (I did test with the number of columns being larger than the number of plots). Good catch.

peterjc:
Fixed, I think. I opted to switch from filling the tile grid by column, to filling by row. Alternatively, I could have let the user set the number of rows?

kiepczi (commented):
Thanks @peterjc for fixing this problem so quick! It looks so much better now. 

Quick question about the `.svgz` files. Can you open them ok on your end? I've tried on both my laptop and MacStudio and both times it shows the following error:
Screenshot 2025-02-19 at 13 57 29

FYI, it doesn't allow me to open this with Safari or Firefox...

peterjc:
No, ``*.svgz`` does not have anything useful associated with it by default on macOS.  If you decompress it, then for me Firefox opens the SVG just fine - e.g.

```console
cat /tmp/sourmash_identity_scatter_1_vs_others.svgz | gunzip > /tmp/sourmash_identity_scatter_1_vs_others.svg
```

Adobe or Inkscape might handle these by default. I think Firefox could handle this with relatively little work - might be worth a look at their issue tracker...

kiepczi:
It's my bad for not realising that .svgz is a zip file. Unzipping it worked fine! I guess this is ready to merge :)

kiepczi (approved):
Ready for merge! Thanks again :)

Issue #244, opened 2025-01-15
Compute ANI from provided alignment


peterjc:
## Summary:

Given *N* genomes and an alignment of *N* sequences with matching names, we can compute pairwise ANI directly, and record the identity matrix from this.

Suitable alignments include full genome alignments from synthetic test cases, down to multi-gene DNA alignments (e.g. core genes or even just one or two marker genes), or perhaps even protein alignments.

## Description:

Possible command line:

```console
> pyani-plus load-alignment --database example.db --create-db --alignment super.fasta genomes/
...
```

We might need optional arguments for:
* how to match the identifiers in the alignment to the FASTA files
* the alignment format (e.g. FASTA, Stockholm, Clustalw - can use Biopython to parse all these)

This should be logged in the DB with a unique configuration in the DB:

* ``configuration.method = "external-alignment"`` or similar (e.g. by filename, filename stem, MD5, or even the FASTA identifier of the first sequence)
* ``configuration.extra = "md5=...;aln=super.fasta"`` or similar can be used to record the filename and checksum of the alignment loaded

This will allow loading multiple alignments for the same set of sequences (e.g. 12 gene alignment versus 10 gene alignment) which would be distinguished by the ``extra`` field.

This will be especially useful with synthetic datasets where the known true-by-construction percentage identity can be loaded and used as a ground-truth in comparative plots.

widdowquinn:
## One input file, or an input directory?

My local test cases for benchmarking are directories with multiple FASTA files - one per aligned contig. This has a happy consequence of avoiding a massive multisequence FASTA file and making subsetting/composition of datasets easier. 

I don't object to being able to cope with a single massive file, but I think it's consistent and helpful to expect the default behaviour of a directory with several FASTA files.

## What's in each file?

The directory with multiple FASTAs makes it eaasier to represent multiple distinct replicons. With a single input FASTA containing multiple sequences we'd need a way (seqID or similar?) to associated replicons from the same individual genome, and also a standard way to label each replicon to ensure the correct sequences were aligned.

This might be easier to manage with multiple FASTA files in a directory.

## Sanity checks for input

For the case with a directory of multiple files, we might want to check for and insist on sequences being presented in the order they have been aligned (e.g. chromosome, plasmid1, plasmid2 in each file). We'd want each sequence that was the same length to be aligned, so padded if necessary.

peterjc:
[...]

PR #292, opened 2025-02-24
Use an internal FASTA parser instead of Biopython (due to import overheads)

peterjc:
Builds on #235 which used lazy-imports in the private CLI to avoid expensive (slow) imports of modules unless directly needed. We currently import one simple function from Biopython:

```python
from Bio.SeqIO.FastaIO import SimpleFastaParser
```

Unfortunately despite this not needing any of the Biopython sequence objects or other parsers, it has a heavy import dependency chain.

Also, this function currently only supports text mode handles, working in unicode. We only really need binary mode and bytes which will be faster for importing large external multiple-sequence alignments, and worth doing anyway given we have to parse every input FASTA file once anyway to extract the total sequence length, and moreso in ANIb.

This PR therefore replaces the use of `SimpleFastaParser` with an equivalent stand-alone implementation, but in binary mode.

This seems to be enough to allow reinstatement of the test intially included on #244 (interupting computating ANI from an external alignment), although I have still seen that fail while importing ``typer`` which is much harder to avoid.

## Type of change

## Action Checklist

- [x] Work on this branch
- [x] Commit branch
- [x] Request a code review

### DIFF
--- a/pyani_plus/utils.py
+++ b/pyani_plus/utils.py
@@ -27,9 +27,66 @@ import os
 import shutil
 import subprocess
 import sys
+from collections.abc import Iterator
 from pathlib import Path
+from typing import IO
 
 from pyani_plus import FASTA_EXTENSIONS
+
+ASCII_GREATER_THAN = ord(">")  # 64
+
+
+def fasta_bytes_iterator(
+    handle: IO[bytes] | gzip.GzipFile,
+) -> Iterator[tuple[bytes, bytes]]:
+    """Parse FASTA file in binary (bytes) mode.
+
+    Yields tuples of (description including identifier, sequence).
+
+    >>> with open("tests/fixtures/viral_example/OP073605.fasta", "rb") as handle:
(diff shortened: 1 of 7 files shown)


pyani_plus/utils.py
```diff
@@ -27,10 +27,67 @@
...
+    # when not opened in universal read lines mode)
+    lines: list[bytes] = []
+    for line in handle:
+        if line[0] == ASCII_GREATER_THAN:  # i.e. >
+            yield title, b"".join(lines).translate(None, b" \t\r\n")
+            lines = []
+            title = line[1:].rstrip()
+            continue
+        lines.append(line.rstrip())
+    yield title, b"".join(lines).translate(None, b" \t\r\n")
```
widdowquinn, line 89: We need to ensure this matches (the `b" \t\r\n"`) with the `yield` statement above - might be worth defining as a variable at the top of the function?

widdowquinn (approved):

Issue #338, opened 2025-04-17
Include Zenodo DOI in CITATIONS file etc


peterjc:
Probably simplest to use https://doi.org/10.5281/zenodo.15005805 which is not version specific, unless nowadays there is an easy workflow to reserve a DOI and commit it to this repository before tagging and making the release files?

widdowquinn:
I believe the umbrella DOI takes you to the latest release, and is the most appropriate.

PR #349, opened 2025-05-02
PyPI, Conda, Zenodo DOI & black style badges

peterjc:
Add PyPI, Conda (these will also be blue once hit v1.0.0), Zenodo DOI (latest version) & black style badges.

cf. https://github.com/peterjc/thapbi-pict/blob/master/README.rst


peterjc:
This would address part of #338

widdowquinn (approved):
Small change, but good. Thanks @peterjc

PR #429, opened 2026-04-29
Add ANIm using minimap2 (ANIminimap2)

peterjc:
See #381 

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [x] This change requires a documentation update
- [ ] This is a documentation update

## Action Checklist

- [x] Work on a single issue/concept (if there are multiple separate issues to address, please use a separate pull request for each)
- [ ] Fork the `pyani-plus` repository under your own account (please [allow write access for repository maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))
- [x] Set up an appropriate development environment (please see `CONTRIBUTING.md` for this repository)
- [x] Create a new branch with a short, descriptive name
- [x] Work on this branch
  - [x] style guidelines have been followed (please see `CONTRIBUTING.md` for this repository)
  - [x] new code is commented, especially in hard-to-understand areas
  - [ ] corresponding changes to documentation have been made
  - [x] tests for the change have been added that demonstrate the fix or feature works
- [x] Test locally with `pytest -v` **non-passing code will not be merged**
- [x] Rebase against `origin/master`
- [x] Check changes with linting/formatting tools before submission (please see `CONTRIBUTING.md` for this repository)
- [x] Commit branch
- [x] Submit pull request, describing the content and intent of the pull request
- [x] Request a code review
- [x] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository

PR #430, opened 2026-05-01
Add lz-ani ANI method

widdowquinn:
This PR adds the lz-ani ANI method, and some syntactic changes to the test suite.

The lz-ani method wraps https://github.com/refresh-bio/LZ-ANI - which is a standalone tool and also the tool performing comparisons in VCLUST (https://github.com/refresh-bio/vclust). In the current implementation, we do not provide access to lz-ani parameters, but we should do this in a future update alongside a `--mode vclust` option to set the specific parameters VCLUST uses for `vclust align`.

The syntactic changes to tests are intended to reduce the number of `ResourceWarning: unclosed database in <sqlite3.Connection object at WHEREVER>` warnings that now get thrown with the coverage checks on python 3.13. The all involve either explicitly closing the sqlite3 database connection, or using a context manager to close the connection automatically. This has not, unfortunately, eliminated the warnings.

## Type of change

- [X] This change requires a documentation update

## Action Checklist

- [X] Work on this branch
- [X] Test locally with `pytest -v` **non-passing code will not be merged**
- [X] Rebase against `origin/master`
- [X] Commit branch
- [X] Request a code review
- [ ] Continue the discussion at [`Pull requests` section](https://github.com/widdowquinn/pyani-plus/pulls) in the `pyan-plus` repository


pyani_plus/methods/lz_ani.py
```diff
@@ -0,0 +1,84 @@
...
+    tuple[str, str, float | None, float | None],
+]:
+    """Parse the lz-ani output files to return a pairwise ANI result.
+
+    :param filename: Path to the lz-ani output file to parse.
+
+    As lz-ani carries out forward and reverse comparisons, we return a tuple of two tuples,
+    one for the forward comparison and one for the reverse comparison. The internal tuples are:
+
+    (query file, subject file, ANI percentage, query coverage)
```
peterjc, line 43: Given the redundancy could return a single tuple of six values?

(query file, subject file, q-vs-s ANI percentage, s-vs-v ANI percentage, query coverage, subject coverage)

But that's a style judgement - and how it gets used may make this form easier...

pyani_plus/private_cli.py
```diff
@@ -54,6 +54,7 @@
     REQ_ARG_TYPE_DATABASE,
     REQ_ARG_TYPE_FASTA_DIR,
 )
+from pyani_plus.utils import stage_file
```
peterjc: All the other methods do this as a lazy import (which can be done more elegantly in Python 3.15 - see eg [this blog](https://techlife.blog/posts/the-story-of-pythons-lazy-imports-why-it-took-three-years-and-two-attempts/)), eg:

```python
from pyani_plus.utils import check_output, stage_file  # noqa: PLC0415
```

This is to speed up the non-computing commands.

pyani_plus/private_cli.py
```diff
@@ -383,7 +384,7 @@ def log_run(  # noqa: PLR0913
         fasta_to_hash=fasta_to_hash,
     )
     # No point caching empty matrices, even partial ones is debatable
-    if run.comparisons().count() == len(fasta_to_hash) ** 2:
+    if run.comparisons().count() == len(fasta_to_hash) ** 2:  # type: ignore[attr-defined]
```
peterjc: Cleaner to pull out the type annotations into their own PR?

pyani_plus/private_cli.py
```diff
@@ -2005,5 +2007,140 @@ def compute_skani(  # noqa: PLR0913, PLR0915
...
+                    "--in-txt",
+                    str(in_txt_path),
+                    "-o",
+                    str(outpath),
+                ],
+            )
+            if not outpath.is_file():
+                msg = f"lz-ani didn't make {outpath}"  # pragma: no cover
+                log_sys_exit(logger, msg)  # pragma: no cover
+            fcomp, _rcomp = parse_lzani(outpath)
```
peterjc: This means doing twice the computation as you throw out the reverse result here, doesn't it?

Better to do a triangle of comparisons like in `compute_external_alignment`, see comments here:

https://github.com/pyani-plus/pyani-plus/blob/08b79ce844d8b33c95703fed1faa3060becc2a58/pyani_plus/private_cli.py#L1820

Actual if statement doing the shortcut is here:

https://github.com/pyani-plus/pyani-plus/blob/v1.0.0/pyani_plus/methods/external_alignment.py#L89

pyani_plus/sql_tables.txt
```diff
@@ -0,0 +1,25 @@
+# CLI
```
peterjc: Was this file checked in by accident?

pyani_plus/tools.py
```diff
@@ -319,6 +319,33 @@ def get_dnadiff(cmd: str | Path = "dnadiff") -> ExternalToolData:
...
+
+    Here the function would return the binary path and "1.2.3" as the version.
+
+    Raises a RuntimeError if the command cannot be found, run, or if the
+    version cannot be inferred - this likely indicates a dramatically
+    different version of the tool with different behaviour.
+    """
+    exe_path, output = _get_path_and_version_output(cmd, ["--version"])
+
+    match = re.search(r"[0-9.]+", output)
```
peterjc: Should you add anchors `^` and `$` to ensure the version is the only text?

tests/test_anib.py
```diff
@@ -129,65 +129,67 @@ def test_running_anib(
...
-        session,
-        run,
-        tmp_json,
-        input_genomes_tiny,
-        hash_to_filename,
-        {},  # not used for ANIb
-        query_hashes=hash_to_length,  # order should not matter!
-        subject_hash=subject_hash,
-    )
+    with db_orm.connect_to_db(logger, tmp_db) as session:
```
peterjc, line 132: As with the typing work, I'd rather you pulled out these context managers into their own pull request.

tests/generate_fixtures/generate_pyani_lzani_matrices.sh
```diff
@@ -0,0 +1,52 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+echo "This is intended to be used with pyani-plust v1+, we have:"
```
peterjc: Typo plust

pyani_plus/public_cli.py
```diff
@@ -712,6 +761,8 @@ def resume(  # noqa: C901, PLR0912, PLR0913, PLR0915
             tool = tools.get_nucmer()
         case "ANIb":
             tool = tools.get_blastn()
+        case "lzani":
```
peterjc: Should this (the name in the DB and thus our reports) not be `LZ-ANI` to match how the authors refer to the tool - https://github.com/refresh-bio/LZ-ANI (which I think would become `lz-ani` at the command line since Typer wants lower case)? Or maybe `LZANI` if the hypen is a pain?

cf `ANIm`, `fastANI`, etc.

peterjc (changes requested):
A couple of things I'd like pull out as separate PR(s) to review and merge first (typing and context managers) - which will make reviewing this much easier, and updating #429 to match a little easier too.

A naming query (why not `LZ-ANI`), and a suggestion for halving the computation.

pyani_plus/private_cli.py
widdowquinn: I hadn't noticed that - apologies! Should be fixed in #ffffe1b (when pushed)

pyani_plus/private_cli.py
widdowquinn: Yes - I think you're right. I'll have to remind myself of the process…

pyani_plus/private_cli.py
widdowquinn: It does. Thanks for the pointer to the shortcut!

pyani_plus/methods/lz_ani.py
widdowquinn: This looks to be tied to the comment about only calculating the triange of comparisons. I'll address both in a commit.

pyani_plus/sql_tables.txt
widdowquinn: Yes - oops! Sorry about that. Removed in #cac10a0 (when pushed)

pyani_plus/tools.py
widdowquinn: Assuming the formatting remains the same (and why wouldn't we?), yes. Corrected in #a8353c7 (when pushed)

tests/test_anib.py
widdowquinn: Will do.

pyani_plus/public_cli.py
widdowquinn: I'll change it for consistency. I think we get to decide what we call our wrapper ;) but I've no objections to changing it.

tests/generate_fixtures/generate_pyani_lzani_matrices.sh
widdowquinn: Fixed in 64652b3

tests/test_anib.py
widdowquinn: Context management was pulled into its own branch and merged with `main` as PR #432

pyani_plus/private_cli.py
widdowquinn: Turns out to have been worth following up. LZ-ANI results are not symmetrical and, if you specify genome order LZ-ANI will ignore you and reorder genomes by itself:

```bash
% lz-ani all2all --multisample-fasta false --in-dir ~/Desktop/2026-07-15_test --in-txt ~/Desktop/2026-07-15_test/intext.txt -o ~/Desktop/2026-07-15_test/test.txt
% cat ~/Desktop/2026-07-15_test/intext.txt
/Users/lpritc/Desktop/2026-07-15_test/MGV-GENOME-0266457.fna
/Users/lpritc/Desktop/2026-07-15_test/MGV-GENOME-0264574.fas
% cat ~/Desktop/2026-07-15_test/test.txt
qidx    ridx    query   reference       tani    gani    ani     qcov    num_alns        len_ratio
1       0       MGV-GENOME-0264574.fas  MGV-GENOME-0266457.fna  0.988142        0.992357        0.995019        0.997325        2       0.9914
0       1       MGV-GENOME-0266457.fna  MGV-GENOME-0264574.fas  0.988142        0.983962        0.994994        0.988912        2       0.9914

% lz-ani all2all --multisample-fasta false --in-dir ~/Desktop/2026-07-15_test --in-txt ~/Desktop/2026-07-15_test/intext_rev.txt -o ~/Desktop/2026-07-15_test/test_rev.txt
% cat ~/Desktop/2026-07-15_test/intext_rev.txt
/Users/lpritc/Desktop/2026-07-15_test/MGV-GENOME-0264574.fas
/Users/lpritc/Desktop/2026-07-15_test/MGV-GENOME-0266457.fna
% cat ~/Desktop/2026-07-15_test/test_rev.txt                                                                                                                                              [5:14:26]
qidx    ridx    query   reference       tani    gani    ani     qcov    num_alns        len_ratio
1       0       MGV-GENOME-0264574.fas  MGV-GENOME-0266457.fna  0.988142        0.992357        0.995019        0.997325        2       0.9914
0       1       MGV-GENOME-0266457.fna  MGV-GENOME-0264574.fas  0.988142        0.983962        0.994994        0.988912        2       0.9914
```

which means that we can't trust `query_hash` and `subject_hash` when adding results to the database 😅

pyani_plus/methods/lz_ani.py
widdowquinn: See notes below on skipping reverse runs

pyani_plus/private_cli.py
widdowquinn: Done.

pyani_plus/private_cli.py
widdowquinn: I've fixed the parsing in [978bb22](https://github.com/pyani-plus/pyani-plus/pull/430/commits/978bb22c744e0abe06486db6598cbaebfd753870), and all tests pass.

However, if I add the trick to ignore redundant runs above at l.2067

```python
            if query_hash < subject_hash or query_hash not in query_hashes:
                # Although LZ-ANI results are not symmetrical in operation, we get forward and
                # reverse results in the same output. This check avoids duplication.
                continue
```

Test resumption fails because Snakemake can't find files:

```text
FAILED tests/test_public_cli.py::test_resume_partial_lzani - SystemExit: Snakemake workflow failed
```

and

```text
INFO     snakemake.logging:__init__.py:1042 Waiting at most 5 seconds for missing files:
/var/folders/9m/72vj0mg57w5_dn3k2qj578500000gn/T/pyani-plus_dxx_ae64/output/LZ-ANI.run_1.column_3.json (missing locally)
ERROR    snakemake.logging:exceptions.py:181 MissingOutputException in rule compute_column in file "/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/pyani_plus/workflows/compute_column.smk", line 28:
Job 0 completed successfully, but some output files are missing. Missing files after 5 seconds. This might be due to filesystem latency. If that is the case, consider to increase the wait time with --latency-wait:
/var/folders/9m/72vj0mg57w5_dn3k2qj578500000gn/T/pyani-plus_dxx_ae64/output/LZ-ANI.run_1.column_3.json (missing locally, parent dir contents: LZ-ANI.run_1.column_1.json)
```

widdowquinn: LZ-ANI is fast enough that we don't lose much efficiency with the redundant comparisons, but this is bugging me.

widdowquinn: I increased the latency-wait setting to 20s and this did not affect the result.

widdowquinn: In the absence of a fix, skipping the shortcut for now.

peterjc:
> Turns out to have been worth following up. LZ-ANI results are not symmetrical and, if you specify genome order LZ-ANI will ignore you and reorder genomes by itself:
>...
> which means that we can't trust query_hash and subject_hash when adding results to the database 😅

Can you reduce that to an even simpler test case, maybe A and B vs C and D, to make it more explicit what is being mixed up here?

Maybe we can force it with naming conventions like qA and qB vs sA and sB (rather than A and B vs A and B), which given we typically stage the files to do A and B and C and ... vs say B, this seems very doable.

widdowquinn:
It's an easy fix to get the appropriate hashes, based on what the output file contains: https://github.com/pyani-plus/pyani-plus/blob/35429f582287ccf7c767d46edc6449c1366fd3db/pyani_plus/private_cli.py#L2125 - I think this means we don't have to make any assumptions about the order (line-by-line) in which LZ-ANI returns the "forward" and "reverse" comparisons, and is the most general solution.

widdowquinn:
> > Turns out to have been worth following up. LZ-ANI results are not symmetrical and, if you specify genome order LZ-ANI will ignore you and reorder genomes by itself:
> > ...
> > which means that we can't trust query_hash and subject_hash when adding results to the database 😅
> 
> Can you reduce that to an even simpler test case, maybe A and B vs C and D, to make it more explicit what is being mixed up here?

I'm attaching a `.zip` file with an example.

[lzani_ordering_test.zip](https://github.com/user-attachments/files/30089208/lzani_ordering_test.zip)

Essentially, to run `lz-ani` on a _specified_ set of genomes (any two or more out of a set, say), you need to pass a plain text file with the full path to each genome file, one genome per line, as the `--in-txt` argument e.g.

```bash
% cat order1.txt
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/A.fasta.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/B.fna.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/C.fas.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/D.fna.gz
```

It doesn't matter which order you pass these sequences in, `lz-ani` will always return the same order of pairwise comparisons. Here, for instance, it is:

```bash
% cat result1.txt
qidx	ridx	query	reference	tani	gani	ani	qcov	num_alns	len_ratio
1	0	D.fna.gz	B.fna.gz	0.350701	0.376535	0.846262	0.444939	9994	0.8499
0	1	B.fna.gz	D.fna.gz	0.350701	0.328745	0.843848	0.389578	11170	0.8499
2	0	C.fas.gz	B.fna.gz	0.338791	0.387806	0.844697	0.459107	8946	0.7380
0	2	B.fna.gz	C.fas.gz	0.338791	0.302615	0.841104	0.359783	10875	0.7380
3	0	A.fasta.gz	B.fna.gz	0.339643	0.389949	0.844750	0.461614	8930	0.7333
0	3	B.fna.gz	A.fasta.gz	0.339643	0.302753	0.841104	0.359947	10872	0.7333
2	1	C.fas.gz	D.fna.gz	0.512457	0.547429	0.868086	0.630616	7849	0.8684
1	2	D.fna.gz	C.fas.gz	0.512457	0.482088	0.866453	0.556392	8415	0.8684
3	1	A.fasta.gz	D.fna.gz	0.513261	0.550207	0.868058	0.633836	7841	0.8628
1	3	D.fna.gz	A.fasta.gz	0.513261	0.481383	0.866442	0.555586	8414	0.8628
3	2	A.fasta.gz	C.fas.gz	0.996938	0.999972	0.999975	0.999997	4	0.9936
2	3	C.fas.gz	A.fasta.gz	0.996938	0.993923	0.999912	0.994011	25	0.9936
```

With any different `--in-txt` ordering, you still get the same result order back. Reversing input order:

```bash
% cat order2.txt
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/D.fna.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/C.fas.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/B.fna.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/A.fasta.gz
```

gives

```bash
% cat result2.txt
qidx	ridx	query	reference	tani	gani	ani	qcov	num_alns	len_ratio
1	0	D.fna.gz	B.fna.gz	0.350701	0.376535	0.846262	0.444939	9994	0.8499
0	1	B.fna.gz	D.fna.gz	0.350701	0.328745	0.843848	0.389578	11170	0.8499
2	0	C.fas.gz	B.fna.gz	0.338791	0.387806	0.844697	0.459107	8946	0.7380
0	2	B.fna.gz	C.fas.gz	0.338791	0.302615	0.841104	0.359783	10875	0.7380
3	0	A.fasta.gz	B.fna.gz	0.339643	0.389949	0.844750	0.461614	8930	0.7333
0	3	B.fna.gz	A.fasta.gz	0.339643	0.302753	0.841104	0.359947	10872	0.7333
2	1	C.fas.gz	D.fna.gz	0.512457	0.547429	0.868086	0.630616	7849	0.8684
1	2	D.fna.gz	C.fas.gz	0.512457	0.482088	0.866453	0.556392	8415	0.8684
3	1	A.fasta.gz	D.fna.gz	0.513261	0.550207	0.868058	0.633836	7841	0.8628
1	3	D.fna.gz	A.fasta.gz	0.513261	0.481383	0.866442	0.555586	8414	0.8628
3	2	A.fasta.gz	C.fas.gz	0.996938	0.999972	0.999975	0.999997	4	0.9936
2	3	C.fas.gz	A.fasta.gz	0.996938	0.993923	0.999912	0.994011	25	0.9936
```

and taking a subset has similar results:

```bash
% cat order5.txt
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/B.fna.gz
/Users/lpritc/Documents/Development/GitHub/pyani-plus/pyani-plus/lzani_ordering_test/genomes/C.fas.gz
```

gives

```bash
% cat result5.txt
qidx	ridx	query	reference	tani	gani	ani	qcov	num_alns	len_ratio
1	0	C.fas.gz	B.fna.gz	0.338791	0.387806	0.844697	0.459107	8946	0.7380
0	1	B.fna.gz	C.fas.gz	0.338791	0.302615	0.841104	0.359783	10875	0.7380
```

> Maybe we can force it with naming conventions like qA and qB vs sA and sB (rather than A and B vs A and B), which given we typically stage the files to do A and B and C and ... vs say B, this seems very doable.

I think it's more robust to get the query/subject names from the output file and use `filename_to_hash()` as above.

peterjc:
Can you get LZ-ANI to do all-vs-one? Specifically give it a set of query genomes and a single subject/reference genome - because that is how our existing per-column snakemake setup works.

I fear not - looking at the docs and https://github.com/refresh-bio/LZ-ANI/blob/master/src/lz-ani.cpp there seems to be only one mode implemented, `all2all`.

In that case we can't really shard the analysis, we just have to call LZ-ANI on the full all-vs-all query (even if we already have most of it pre-computed). If it is fast it won't really matter.

It may be worth proposing they add an `all2one` mode with a suggested API?

widdowquinn:
I did look into all-v-one but no, it's not possible. `lz-ani` looks to be optimised to the authors' needs, rather than ours. We can't shard what `lz-ani` does in the way we do for `sourmash`, but I think it would - in principle, be possible to divide a 2x2 matrix of comparisons into "squares." Whether that overhead is worth it, I don't know. `lz-ani` is _very_ fast, and we have a working implementation, so we can raise it as an issue and decide whether to switch on the basis of performance when it's written, maybe?

We _could_ ask the authors to do `all2one` but then - if they agreed - we'd be waiting for their implementation and we'd be waiting until we updated ours to use it. And, in the meantime, we _do_ have a working implementation…

peterjc:
Got it - so we make roughly N^2 separate calls (or did you manage to halve it now?) and for each element of the matrix (so self-vs-self aside) we've actually giving LZ-ANI two different genomes in all-vs-all mode (and so should get back both A vs B and B vs A numbers). Assuming it itself shortcuts self-vs-self there is no wasted computation on ANI values (just the overheads of lots of CLI tool calls).

widdowquinn:
Yes we're making n^2 calls to `lz-ani`, each call being a pairwise comparison between two genomes. `lz-ani all2all` gives you forward and reverse results (A vs B, B vs A) and I couldn't see a way to only have it return a single call.

In principle we could use the trick above (as with the exact comparisons) to halve the number of calls by skipping the reverse comparison _call_, but I still can't work out why employing that causes the resume partial lz-ani test to fail, so we're duplicating results. We're making n^2 calls and getting 2*(n^2) comparisons done 😭

Any insight you have as to why the resume partial run tests fail would be most welcome!

pyani_plus/private_cli.py
```diff
@@ -2170,5 +2171,160 @@ def compute_skani(  # noqa: PLR0913, PLR0915
...
+                    "query_hash": filename_to_hash[rcomp[0]],
+                    "subject_hash": filename_to_hash[rcomp[1]],
+                    "identity": rcomp[2],
+                    "cov_query": rcomp[3],
+                    "configuration_id": config_id,
+                    "uname_system": uname_system,
+                    "uname_release": uname_release,
+                    "uname_machine": uname_machine,
+                }
+            )
```
peterjc, line 2286: You can avoid the code repetition with two explicit `db_entries.append(...)` with a loop over the two results (forward and reverse).

pyani_plus/private_cli.py
```diff
@@ -2170,5 +2171,160 @@ def compute_skani(  # noqa: PLR0913, PLR0915
...
+
+            new += 1
```
peterjc: Probably `new += 2` ~~but I suspect we can remove `new` as a vestige of an older iteration of the batched JSON output.~~

peterjc: The final line skips the JSON output if new is false (zero), so the actual count isn't used.

pyani_plus/methods/lz_ani.py
peterjc: Returning two tuples will simplify recording the forward and reverse results (A vs B, and  B vs A).

peterjc:
I'm not seeing anything obvious about why skipping half the matrix wouldn't work (just staring at the code).

In practice with this method, if A vs B is missing, then B vs A is also missing. When we work out what data is missing and thus which columns (all-vs-?) jobs are needed, I think the current logic will start both an all-vs-A job and an all-vs-B job - even though only one of those is really needed (will give both missing numbers) as the other will skip the comparison.

It could be the compute all-vs-column for LZ-ANI does not fail gracefully if there seems to be nothing to do?

pyani_plus/private_cli.py
widdowquinn: I've removed the repetition by using `db_entries.extend()`

pyani_plus/private_cli.py
widdowquinn: I've changed to `new += 2` even though the count isn't used explicitly.

